PR #46, opened 2013-08-19
errorFormatter called with the request as this / bracket notation/ array index

itoche:
The idea is to call errorFormatter() with the request as this to allow an errorFormatter custom implementation to make use of other information/functions stored on the request for the formatting of the error.

My practical case is to internationalize the message thanks to the user's locale stored along each request.

I also added:
* support for array index when defining the param path like 
    req.checkBody(['users', 0, 'name'], 'error message').notEmpty()
* support for bracket notation when defining the param path like
    req.checkBody('users[0][name]', 'error message').notEmpty()

Let me know what you think of this.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -45,7 +45,7 @@ var expressValidator = function(options) {
     return function(param, fail_msg) {
 
       var value;
-
+      var paramPath = param;
       // If param is not an array, then split by dot notation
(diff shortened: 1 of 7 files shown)


rustybailey:
This PR can't be merged. Also, it would be preferable if you could open a PR for each separate enhancement. Please open a new PR if this request is still relevant.

PR #51, opened 2013-09-04
Expose node-validator instance as module.exports.validator so it can be ...

### LINKED ISSUE #50: Expose node-validator instance for easier customization
To customize things like defaultError, we need the node-validator instance.

jerryorr:
...customized

Fixes #50


rustybailey:
node-validator was exposed in https://github.com/ctavan/express-validator/pull/70

PR #71, opened 2014-01-28
Support for validation against req.files

dapids:
Hey Christoph,

First of all thanks a lot for the nice job you've made so far with express-validator. ;)

I've just added the the possibility to validate against req.files. I wanted to ask you how I should behave about the tests. I've seen that you're using node-request for testing the http requests, but I'm not sure whether it supports file uploading. In that case I'd be happy to write some tests, too.

Best,
David


rustybailey:
Closing as this was partially implemented and @dapids hasn't responded in almost a year. Please open a new PR if you would like to add your change to `updateParams`.

PR #82, opened 2014-04-04
Added the option to clear errors after retrieving them

flore2003:
When using express-validator in multiple chained middlewares, it is useful to optionally clear the validation errors after retrieving them, so that errors from the last middleware don't interfere with the current middleware.

I have added an optional second parameter to the expressValidator.validationErrors method, that specifies whether the current errors should be cleared after retrieving them.

Example usage:
```js
var firstMiddleware = function(req, res, next) {
    req.checkParams("something", "No valid something specified").isInt();
    var errros = req.validationErrors(false, true); // this clears the errors
    if(errors) {
        // do stuff to handle validation error
        return next();
    }
    next();
};

var secondMiddleware = function(req, res, next) {
    req.checkQuery("p", "No valid p specified").notEmpty().isInt();
    var errors = req.validationErrors();
    // without this, errors would contain all the errors that have occurred in earlier middleware,
    // even though they have been handled
    if(errors) {
        // return some kind of error to the user
    }
    // do useful stuff
};

app.get("/:something", firstMiddleware, secondMiddleware);
```

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -196,7 +196,7 @@ var expressValidator = function(options) {
       }
       if(clear) {
       // Clear errors after retrieving them
-        req._validationErrors = [];
+        req._validationErrors = undefined;
       }
       return errors;
     }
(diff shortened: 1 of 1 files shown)


rustybailey:

PR #109, opened 2014-12-01
Add methodName to error handler in order be able to know witch validatio...

erubio:
Hey Christoph,

First of all good job with that middleware.

I think could be usefull know witch validation has failed in a chain of falidations, in order to know that i modify  the code to send this information to the error handler, and in that case we can know this information or even change the message of the error to know what really happened.

What do you think about this change?

Thanks a lot!


rustybailey:
This is too far out of sync. Please rebase and reopen if this is still needed.

PR #117, opened 2015-02-06
replaced request.param(item) with request.body[item]

properjon:
request.param() is deprecated as of 4.11


SamVerschueren:
Doesn't cover all the use cases. It's only possible to sanitize properties in the body. But what with properties in the querystring?

```req.check``` should be kept the way it was. You can use ```req.checkBody``` instead to get rid of the deprecation warning. It should be kept this way for backwards compatibility.

ctavan:
published as 2.9.0

PR #134, opened 2015-05-01
Remove old unit tests in favor of Mocha unit tests

### LINKED ISSUE #133: Move unit tests to Mocha
- Move tests to use Mocha/Chai
- Make tests more cross-platform compatible
    - Remove Makefile
    - Change `npm test` to use mocha (recursive)

rustybailey:
Changes made:
- All old tests removed and replaced with Mocha unit tests
- Makefile removed and `npm test` is changed to use `mocha ./test`

Notes:
- I didn't add any new tests. All I did was convert the old tests to Mocha so that the descriptions are more verbose and it's easy to run the tests on any machine.
- I changed the way the `app.js` helper worked
     - Because of that, I had to do a weird thing where I cleared require's cache entry for that file before each set of tests. Otherwise, there was no way to change the validation middleware for each test file. 
     - Maybe there's a better way to do this, but after a lot of research, that was the best way I could find.
     - Ideally, it would be nice to dynamically change the middleware at runtime so each test could change the validation function if it needed to, but I couldn't find an easy way to do that.

**Any feedback is much appreciated**

This closes #133.


ctavan:
Nice! Would you mind rebasing?

rustybailey:
Sure thing. I won't get to this until later tonight, but when I do, I can merge, bump the version and publish.

PR #137, opened 2015-05-25
added custom santizers

uhlryk:
Hi
In my project I needed the ability to add custom santizers. For example you have normalizeEmail but I need normalizePhone. So i did it. Now you can in options add customSantizers, it is very similar to customValidators (it is refactored customValidators). I have not found tests for customValidators, so I did not make tests for this method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,26 @@ Use them with their validator name:
 req.checkBody('users', 'Users must be an array').isArray();
 req.checkQuery('time', 'Time must be an integer great than or equal to 5').isInt().gte(5)
 ```
+####`customSanitizers`
+_{ "sanitizerName": function(value, [additional arguments]), ... }_
+
+The `customSanitizers` option can be used to add additional sanitizers methods as needed. This option should be an `Object` defining the sanitizer names and associated functions.
+
+Define your custom sanitizers:
+
+```javascript
+app.use(expressValidator({
+ customSanitizers: {
+    toStingLowerCase: function(value) {
+        return (""+value).toLowerCase();
+    },
+ }
+}));
+```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -124,6 +124,26 @@ Use them with their validator name:
...
+_{ "sanitizerName": function(value, [additional arguments]), ... }_
+
+The `customSanitizers` option can be used to add additional sanitizers methods as needed. This option should be an `Object` defining the sanitizer names and associated functions.
+
+Define your custom sanitizers:
+
+```javascript
+app.use(expressValidator({
+ customSanitizers: {
+    toStingLowerCase: function(value) {
```
rustybailey: I know this is just an example method, but I would prefer for "String" to be spelled correctly. Same for the instance of "toStingLowerCase" below.

uhlryk: yes, you are, correct, i made it fast, in weekend i will correct this, make tests and change to single commit, and then pull request.

rustybailey: No worries! I doesn't matter if this is multiple commits or not, so just push your new changes to this PR. If you could, ping me on here when you're finished making changes since I don't get notified when new commits are added. Thanks again!

uhlryk: Hi
I just commit.
I like your module, it is very useful. I will use it my next projects and
if find something to improve i help.
I thing about chain sanitizers, or even chain sanitizers with validators in
one chain. It is not important feature but maybe useful.
Or if somebody make custom validator which use events then promises would
be helpful. For example you validate if email is in db, result of db you
get with promise.

2015-05-28 20:17 GMT+02:00 Rusty Bailey <notifications@github.com>:

PR #144, opened 2015-06-16
Provide a default options value

### LINKED ISSUE #143: Inconstency between readme / doc and lib
doc says constructor `options` is optional but if we do not provide an object we got this error

```
TypeError: Cannot read property 'customValidators' of undefined
    at expressValidator (/.../node_modules/express-validator/lib/express_validator.js:101:20)
```

It is because this [call](https://github.com/ctavan/express-validator/blob/master/lib/express_validator.js#L72) does not provide a default `options` value. `_.defaults` will return null if `options` is null, even with a correct `defaults` object.

JSteunou:
should fix #143

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -57,6 +57,7 @@ function Sanitizer(param, req, locations) {
  */
 
 var expressValidator = function(options) {
+  options = options || {};
   var defaults = {
     customValidators: {},
     customSanitizers: {},


rustybailey:
Thanks, this was something that was lost during the refactor.

rustybailey:
This fix has been published to npm as 2.12.1.

JSteunou:
wow fast! thank you.

rustybailey:
Ha ha. Just happened to sync up. I had just published the minor version you found a bug in.

PR #151, opened 2015-07-10
Support override message in chain

bolasblack:
Like `req.checkParam('user_id', 'required').isInt({message: 'invalid'})`


rustybailey:
This was added by https://github.com/ctavan/express-validator/pull/159 and is in version 2.15.0

PR #155, opened 2015-08-10
Don't call validators for optional parameters

ikokostya:


rustybailey:
Good code clarification. Thanks for the contribution.

rustybailey:
This is now on npm as 2.14.1.

ikokostya:
Thanks for quick reply.

PR #157, opened 2015-08-11
Fixed typo with the direction of how to use custom sanitizers.

phuson:


rustybailey:
Good catch! Thanks!

rustybailey:
This fix is in 2.14.2

phuson:
Thanks!

PR #158, opened 2015-08-13
Fix a typo in README

supasate:
Fix a typo of a sanitizer example in README to match the returned value after calling trim().

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ req.checkBody('email').optional().isEmail();
 ```javascript
 
 req.body.comment = 'a <span>comment</span>';
-req.body.comment.username = '    user    ';
+req.body.comment.username = '   a user    ';
 
 req.sanitize('comment').escape(); // returns 'a &lt;span&gt;comment&lt;/span&gt;'
 req.sanitize('comment.user').trim(); // returns 'a user'


README.md
```diff
@@ -238,7 +238,7 @@ req.checkBody('email').optional().isEmail();
 ```javascript
 
 req.body.comment = 'a <span>comment</span>';
-req.body.comment.username = '    user    ';
+req.body.comment.username = '   a user    ';
```
rustybailey: Good catch, but it made me realize that the whole example doesn't really make sense considering `username` couldn't be a property on a string (`comment`) and the example wasn't consistent in using `username`/`user`. Can you change that javascript section to:

```javascript

req.body.comment = 'a <span>comment</span>';
req.body.username = '   a user    ';

req.sanitize('comment').escape(); // returns 'a &lt;span&gt;comment&lt;/span&gt;'
req.sanitize('username').trim(); // returns 'a user'

console.log(req.body.comment); // 'a &lt;span&gt;comment&lt;/span&gt;'
console.log(req.body.username); // 'a user'

```

rustybailey: I went ahead and made all these changes myself.

Issue #126, opened 2015-04-06
Custom Error Message For Chain Validation


slier81:
```
req.checkBody('age', 'Invalid age').notEmpty().isInt();
var errors = req.validationErrors(true); 
```

This will always return `Invalid age`

How can I return `Age is empty`, when input does not pass `notEmpty()` method, and return `Age should be an integer` if it does not pass `isInt()` method?

Maybe something like this:
```
req.checkBody('age').notEmpty("Age is empty").isInt("Age should be an integer");
```

I dont want user always get `Invalid age` when he/she does not supply input for age

rustybailey:
We can't necessarily do it like you're suggesting because those validator functions may or may not take inputs, so that would conflict. My first idea is that we could do an object in order to map messages to certain functions, like so:
``` js
req.assert('age', { notEmpty: 'Age is empty', isInt: 'Age should be an integer' }).notEmpty().isInt();
```
In the current state, I think we assume you're sending a string, but you could technically send anything in that second param, so giving objects a specific functionality could be potentially breaking if anyone is using something in that second param besides strings. So if we decided to implement it this way, it would 1) be a breaking change; 2) be limiting if someone wanted to send something other than a string for the error message.

Thoughts, @ctavan?

ctavan:
We could make all arguments mandatory in case you want to supply an error message. For validation functions with optional arguments we would then require to pass optional arguments as explicit `undefined`.

So e.g. for `isISBN(str [, version])` the call signature with error message would be:

```js
req.checkBody('isbn').isISBN(undefined, 'Not a valid ISBN');
```

What do you think?

slier81:
@rustybailey @ctavan maps work great but does not clean to me. as suggested by @ctavan supplying error message on the function itself look much cleaner.. i think the function itself should take a maps.

Something like this:

```
req.checkBody('isbn').isISBN({
  vesion : 'blah',
  message: 'not valid message'
});
```

rustybailey:
Personally, I don't like the idea of putting an additional param or an object into the actual validation method. Part of me feels like if we're going to this much trouble to think of a way to shove in multiple error messages for chained validation, we should just recommend that people use two different functions if they want different error messages.

chrissinclair:
Did anything much come of this? I've just started playing around with the library, and it's the first real issue I've hit. If you're still taking suggestions then what about something along the lines of 
```
req.check('postcode')
    .isLength(6, 8).withMessage('Invalid length for postcode')
[...]

PR #159, opened 2015-08-16
#126 - Added withMessage method to provide a message per validation

chrissinclair:
Currently, when validations are chained, only a single message to be used for all errors in the chain can be provided. This PR adds a method `.withMessage(message)`, to be called immediately after the validation in the chain, that provides the new message via the errorFormatter. 

In this example:
```javascript
req.check('param', 'Default Message')
    .notEmpty()
    .isInt().withMessage('Custom message')
```

if 'param' is empty, the message 'Default Message' will be used, but if it is provided and is not an integer, 'Custom Message' will be used instead. 

The implementation is simple enough - just track the param & value of the last error generated and then when `withMessage` is called, pop the last error and re-generate it with the new message. 

Tests provided, any feedback welcome.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -23,7 +23,7 @@ function ValidatorChain(param, failMsg, req, location, options) {
   this.validationErrors = [];
   this.failMsg = failMsg;
   this.req = req;
-  this.lastError = undefined; // used by withMessage to get the values of the last error
+  this.lastError = null; // used by withMessage to get the values of the last error
   return this;
 }
 
@@ -100,13 +100,13 @@ var expressValidator = function(options) {
   };
 
   ValidatorChain.prototype.withMessage = function(message) {
-    if (!!this.lastError) {
-      var error = this.validationErrors.pop();
+    if (this.lastError) {
+      this.validationErrors.pop();
       this.req._validationErrors.pop();
-      error = this.errorFormatter(this.lastError.param, message, this.lastError.value);
+      var error = this.errorFormatter(this.lastError.param, message, this.lastError.value);
       this.validationErrors.push(error);
       this.req._validationErrors.push(error);
-      this.lastError = undefined;
+      this.lastError = null;
     }
 
     return this;
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -23,7 +23,7 @@ function ValidatorChain(param, failMsg, req, location, options) {
   this.validationErrors = [];
   this.failMsg = failMsg;
   this.req = req;
-
+  this.lastError = undefined; // used by withMessage to get the values of the last error
   return this;
 }
 
@@ -94,6 +94,19 @@ var expressValidator = function(options) {
   ValidatorChain.prototype.optional = function() {
     if (this.value === undefined) {
       this.skipValidating = true;
+    }
+
+    return this;
+  };
+
+  ValidatorChain.prototype.withMessage = function(message) {
+    if (!!this.lastError) {
+      var error = this.validationErrors.pop();
+      this.req._validationErrors.pop();
+      error = this.errorFormatter(this.lastError.param, message, this.lastError.value);
+      this.validationErrors.push(error);
+      this.req._validationErrors.push(error);
+      this.lastError = undefined;
     }
 
     return this;
@@ -188,6 +201,7 @@ function makeValidator(methodName, container) {
       var error = this.errorFormatter(this.param, this.failMsg || 'Invalid value', this.value);
       this.validationErrors.push(error);
       this.req._validationErrors.push(error);
+      this.lastError = { param: this.param, value: this.value };
     }
 
     return this;
--- a/test/withMessageTest.js
+++ b/test/withMessageTest.js
@@ -1,0 +1,75 @@
+var chai = require('chai');
+var expect = chai.expect;
+var request = require('supertest');
+
+var app;
+var errorMessage = 'Default error message';
+var mustBeTwoDigitsMessage = 'testparam must have two digits';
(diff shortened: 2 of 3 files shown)


rustybailey:
Awesome! Thanks for the contribution! I'm busy for most of the day, so I won't be able to give it a thorough review until tonight or tomorrow. However, one thing I did notice is that you should include some documentation for this feature in the README. Thanks again!

chrissinclair:
Ha - yeah that would have been useful. Thanks for having a look, let me know if you see anything else that needs a tweak. 

Cheers :)

lib/express_validator.js
```diff
@@ -23,7 +23,7 @@ function ValidatorChain(param, failMsg, req, location, options) {
   this.validationErrors = [];
   this.failMsg = failMsg;
   this.req = req;
-
+  this.lastError = undefined; // used by withMessage to get the values of the last error
```
rustybailey: Let's initialize this as `null`. It's kind of paradoxical to define something as `undefined`.

lib/express_validator.js
```diff
@@ -99,6 +99,19 @@ var expressValidator = function(options) {
     return this;
   };
 
+  ValidatorChain.prototype.withMessage = function(message) {
+    if (!!this.lastError) {
```
rustybailey: No need for the double-negative boolean conversion since the `if` will already evaluate it as boolean.

lib/express_validator.js
```diff
@@ -99,6 +99,19 @@ var expressValidator = function(options) {
...
+      var error = this.validationErrors.pop();
```
rustybailey: Is there a reason we define error here? I would think it only needs to be defined the second time.

lib/express_validator.js
```diff
@@ -99,6 +99,19 @@ var expressValidator = function(options) {
...
+      this.req._validationErrors.pop();
+      error = this.errorFormatter(this.lastError.param, message, this.lastError.value);
+      this.validationErrors.push(error);
+      this.req._validationErrors.push(error);
+      this.lastError = undefined;
```
rustybailey: `this.lastError = null;` for same reason as above

test/withMessageTest.js
```diff
@@ -0,0 +1,75 @@
...
+  delete require.cache[require.resolve('./helpers/app')];
+  app = require('./helpers/app')(validation);
+});
+
+describe('#withMessage()', function() {
+    it('should return one error per validation when param does not validate as two digit int, with custom message', function(done) {
+      getRoute('/test', fail([mustBeIntegerMessage, errorMessage, mustBeTwoDigitsMessage]), 3, done);
+    });
+
+    it('should return three errors when param is missing, with default message for the first and custom messages for the rest', function(done) {
```
rustybailey: Should say "four errors"

test/withMessageTest.js
```diff
@@ -0,0 +1,75 @@
...
+      getRoute('/', fail([errorMessage, mustBeIntegerMessage, errorMessage, mustBeTwoDigitsMessage]), 4, done);
+    });
+
+    it('should return a success when param validates', function(done) {
+      getRoute('/42', pass, null, done);
+    });
+
+    it('should return a fail with a single custom message when failing one validiation', function(done) {
```
rustybailey: "validation" typo

test/withMessageTest.js
```diff
@@ -0,0 +1,75 @@
...
+var request = require('supertest');
+
+var app;
+var errorMessage = 'Default error message';
+var mustBeTwoDigitsMessage = 'testparam must have two digits';
+var mustBeIntegerMessage = 'testparam must be an integer';
+
+function validation(req, res) {
+  req.checkParams('testparam', errorMessage)
+    .notEmpty()
```
rustybailey: Maybe have another comment here that says `// with default message` like `isInt()` below, just so it's super clear.

rustybailey:
Had a few comments for your code -- just make sure the tests are clear in what you're testing. Please leave a comment when you've fixed everything so that I know it's good for a final review/merge.

*Note: This PR should resolve https://github.com/ctavan/express-validator/issues/126*

chrissinclair:
Ok, I've updated that. Not sure why the CI build couldn't complete though - I'm getting no errors running `npm run travis-build` locally, and looking at the details it's claiming npm install has failed - pretty sure I've not touched anything that would give that cause to break. Any ideas?

rustybailey:
Not sure why it errored out on just a WARN. I re-ran the build and it finished fine. I will merge this in and update by the end of the day.

chrissinclair:
Weird - glad it's working though! Thanks for having a look :)

chrissinclair:
Hmm - had a thought this morning I was just able to check out. Looks like it's not quite right just yet:

```javascript
it('should update the error message only if the preceeding validation was the one to fail', function(){
      var validator = require('../lib/express_validator')();
      var req = {
        body: {
          testParam: 'abc'
        }
      };
      
      validator(req, {}, function() {});
      req.check('testParam', 'Default Error Message')
        .isInt() // should produce 'Default Error Message'
        .isLength(2).withMessage('Custom Error Message');
        
      expect(req.validationErrors()).to.deep.equal([
        {param: 'testParam', msg: 'Default Error Message', value: 'abc'}
      ]);
      // fails as the message is 'Custom Error Message'
    });
```

This test will fail as we're updating the error for the wrong test. I think the fix should be simple enough (clear the lastError field on a successful validation) - do you want this as a new PR?

rustybailey:
I would think this test would fail because there would be two items in the validation errors array. Either way, if you think there is a hole in the PR, please open a new one with the fix and tests that cover that scenario.

PR #160, opened 2015-08-18
Fixed bug where custom messages would be applied incorrectly

chrissinclair:
Currently, if in a chained validation one validation fails, and no withMessage is provided, but all further validations pass and a with message is provided, the original error will be given the custom message incorrectly. 

This PR fixes this issue, by ensuring that if a validation passes no calls to withMessage further down the chain will attempt to update the error message. 

Test included, feedback welcome.


rustybailey:
Fix on 2.15.1

Issue #122, opened 2015-03-01
Calling an async method from a custom validator


stonebraker:
I'm attempting to write an isUnique custom validator which requires an async call to mongo. However, custom validators are implemented as synchronous methods without callbacks. Can you recommend a technique for accomplishing this?

sluchznak:
Hi, i have the same problem. Please, notify me, when you find solution

lorenz:
+1

jp1987:
This would be great! Really need a way to integrate mongo in the validation.

rustybailey:
This feature should now be available in 2.17.0. Please check the documentation for [async validation](https://github.com/ctavan/express-validator#asynchronous-validation) and open a new issue if that functionality doesn't work properly.

Issue #94, opened 2014-08-15
Is it possible to add my own validator, which would return a Promise?


longuniquename:
My use case is to check if an email is unique across DB table. Pretty common issue. I use [Bookshelf](http://bookshelfjs.org/) to fetch my data. And it's asynchronous! How can I wrap this check into custom validator?

sgarbesi:
+1

joelhdzfdz:
I don't think returning a promise would be a good idea as it would break the syntactic sugar of concatenation, however providing

joelhdzfdz:
I don't think returning a promise would be good as it would break the syntatic sugar, however providing a success and failure method to the custom validators arguments would suffice to accomplish any kind of async validation such as checking wether a record exists in a database. This would look like 

function(value,success,failure){
    models.Country.find({where : {code : value}}).then(function(){
        success();
    },function(){
        failure();
    });
};

joelhdzfdz:
I don't think returning a promise would be good as it would break the syntatic sugar, however providing a success and failure method to the custom validators arguments would suffice to accomplish any kind of async validation such as checking wether a record exists in a database. This would look like 
```
function(value,success,failure){
    models.Country.find({where : {code : value}}).then(function(){
        success();
    },function(){
        failure();
    });
};
```

firehawk895:
@joelhdzfdz I didn't quite understand the solution.

renegare:
This would be great given the growing number of promise based libraries.

rustybailey:
This feature should now be available in 2.17.0. Please check the documentation for [async validation](https://github.com/ctavan/express-validator#asynchronous-validation) and open a new issue if that functionality doesn't work properly.

renegare:
have not checked but +1, will do though ;)

PR #164, opened 2015-08-26
Async Validation

Jakeii:
Adds the functionality to be able to return a promise from a custom validator.

Requires that `req.asyncValidationErrors()` is used which will return a promise, rather than `req.validationErrors()` otherwise async validators may not have completed. I added a console warning if you use an async validator and then use `validationErrors`, as it could be a common mistake...

`asyncValidationErrors` will resolve if there are no errors and will reject with any (a)synchronous validation errors!

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -1,5 +1,6 @@
 var validator = require('validator');
 var _ = require('lodash');
+var Promise = require('bluebird');
 
 // validators and sanitizers not prefixed with is/to
 var additionalValidators = ['contains', 'equals', 'matches'];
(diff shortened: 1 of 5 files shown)


rustybailey:
This is awesome! Sorry, but I have been busy the past couple days or your other PR would've already been merged. I will make a point to look over this tonight.

lib/express_validator.js
```diff
@@ -124,7 +125,12 @@ var expressValidator = function(options) {
     var locations = ['body', 'params', 'query'];
 
     req._validationErrors = [];
-    req.validationErrors = function(mapped) {
+    req._asyncValidationErrors = [];
+    req.validationErrors = function(mapped, promisesResolved) {
+      if (!promisesResolved && req._asyncValidationErrors.length > 0) {
+        console.warn('WARNING: You have asynchronus validators but you have not used asyncValidateErrors to check for errors.');
```
rustybailey: Typo: asynchronus -> asynchronous

rustybailey: Also, not sure what the correct answer is here... but should we only do a "warn" or should we also throw an Error? What are the consequences of using asynchronous validators without `asyncValidateErrors`?

rustybailey:
Just one small comment, but besides that, everything looks good from a code perspective. You'll need to resolve conflicts with master since I merged in your other PR.

*Note: This should fix https://github.com/ctavan/express-validator/issues/122 and https://github.com/ctavan/express-validator/issues/94* :clap:

lib/express_validator.js
Jakeii: Async validation errors simply won't show for `validationErrors()`.

rustybailey:
This feature is now available in 2.17.0

Issue #165, opened 2015-08-28
the two way of nested values' mappedErrors are different?


alexgzhou:
req.assert(['user', 'email'], 'not valid email').optional().isEmail();
req.assert('user.mobile', 'not valid mobile').optional().isMobilePhone('zh-CN');

In the previous express-validator version, if i use the two req.assert above to check the nested params, I got the same mappedErrors as mappedErrors['user.email'] and mappedErrors['user.mobile']; 
But now I get mappedErrors['user,email'] and mappedErrors['user.mobile'], **notice** the separator is a **comma** in the 'user,email', not a dot, is it on purpose?

And when I check three layers nested values like below, it does not work:
req.assert('user.personalInfo.idNo', 'not valid ID No.').optional().len(6, 20).isAlphanumeric();

And below is working, but again, I got a mappedError like mappedError['user,personalInfo.idNo'], **notice** the first separator is a **comma**, not a dot:
req.assert(['user','personalInfo.idNo'], 'not valid ID No.').optional().len(6, 20).isAlphanumeric();

Is it true that only two values nested is OK with the syntax of 'dot-notation'?

rustybailey:
Thanks for the reporting this. What was the last version you used that worked?

rustybailey:
@Jakeii or @chrissinclair: would either of you be able to look into this?

rustybailey:
After doing some investigation, this errors seems to have occurred because of the giant code refactor a couple months ago. Essentially, in [v2.11.0](https://github.com/ctavan/express-validator/tree/v2.11.0) (before the refactor), if you ran the test `nestedInputTest.js`, the errors would output as:
``` js
[ { param: 'user.fields.email', msg: 'not empty', value: '' },
  { param: 'user.fields.email', msg: 'not empty', value: '' },
  { param: 'user.fields.email', msg: 'valid email required', value: '' } ]
```
However, in [v2.12.0](https://github.com/ctavan/express-validator/tree/v2.12.0) (after the refactor), they output as:
``` js
[ { param: [ 'user', 'fields', 'email' ],
    msg: 'not empty',
    value: '' },
  { param: 'user.fields.email', msg: 'not empty', value: '' },
  { param: [ 'user', 'fields', 'email' ],
    msg: 'valid email required',
    value: '' } ]
```
Notice how it doesn't convert `[ 'user', 'fields', 'email' ]` to `'user.fields.email'` in the param field.

So we need to:
[...]

PR #166, opened 2015-08-28
Fix the output of nested params in error object.

rustybailey:
This is a fix for https://github.com/ctavan/express-validator/issues/165.

I wrapped `errorFormatter()` in a function called `formatParamOutput` that formats the param for output. Also exported this function so that I could easily test it. In the future we could possibly add other functions to the utils object for testing.

Would appreciate a code review by @chrissinclair or @Jakeii. Just leave a comment if you've finished reviewing or have any feedback.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -104,7 +104,7 @@ var expressValidator = function(options) {
     if (this.lastError) {
       this.validationErrors.pop();
       this.req._validationErrors.pop();
-      var error = this.errorFormatter(this.lastError.param, message, this.lastError.value);
+      var error = formatErrors.call(this, this.lastError.param, message, this.lastError.value);
       this.validationErrors.push(error);
       this.req._validationErrors.push(error);
       this.lastError = null;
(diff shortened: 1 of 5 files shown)


lib/express_validator.js
```diff
@@ -340,5 +339,47 @@ function locate(req, name) {
...
+  var formattedParam = formatParamOutput(param);
+
+  return this.errorFormatter(formattedParam, msg, value);
+}
+
+// Convert nested params as array into string for output
+// Ex: ['users', '0', 'fields', 'email'] to 'users[0].fields.email'
+function formatParamOutput(param) {
+  if (Array.isArray(param)) {
+    param = param.reduce(function(prev, curr) {
```
Jakeii: How about using a regex here?
````javascript
param = param.join('.').replace(/\.([0-9]+)\./, '[$1].');
````

rustybailey: That regex only replaced the first number it found (no global flag) and it didn't replace it if the number it found was at the end of the string (since it wouldn't match `/\.([0-9]+)\./`). I added a few tests to account for these scenarios in case you wanted to try tweaking your regex suggestion.

rustybailey:
Going to go ahead and merge this so we can get this fix out there.

PR #183, opened 2015-10-28
Optional() options, sanitizeHeaders(), and documentation

### LINKED ISSUE #120: Allow "optional" to test for truthy
Currently `optional()` tests whether the expected parameter is defined or not. I'd be great if, even when defined, it be classified as optional if it evaluates to falsey. In other words, these tests should actually `pass`:

https://github.com/ctavan/express-validator/blob/master/test/optional.js#L39-L40

I suspect it'd be a more general resolution to #113 as well.

While I can create a custom validator that tests whether a value is truthy or falsey, it won't short-circuit the remaining validation declarations. So:

```javascript
isReallyOptional: function(value) {
    return !value;
  }
```

Being used like:

```javascript
req.checkQuery('addr', 'if specified, must be three or more characters')
    .isReallyOptional().len(3);
```

Passes my `isReallyOptional` check but then, of course, fails the `len` check.

Current work around is to make a single custom validator that does all of my checks (rather than chaining them):

```javascript
'use strict';
var validator = require('validator');

var Validators = {
  optionalOrLen: function(value, minLength) {
    return !value || validator.isLength(value, minLength);
  }
};

module.exports = Validators;
```

### LINKED ISSUE #177: sanitizeHeaders missing?
Is there any reason `sanitizeHeaders` does not exist? I am parsing a custom header from a string into a custom type, but I keep getting `undefined is not a function` when I call `req.sanitizeHeaders`.

### LINKED ISSUE #182: How to call sanitizers with extra arguments?
Hello!

There are no examples of using sanitizers with extra parameters in the docs, e.g.: `blacklist`.

Question: What is the proper way to call them?
Feature Request: Please update the docs accordingly.

Thank you!

rustybailey:
- Added options arguments to `optional()` to allow for checking for falsy values. (fix #120)
- Added docs for sanitizer arguments (fix #182)
- Added `sanitizeHeaders()` with tests and docs. (fix #177)
- Update validator to 4.2.x
- Added Node v4.0 and v4.1 to Travis.

@ctavan ... Thoughts? I attempted to add that feature to `optional()` while still being backwards-compatible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: node_js
 node_js:
   - "0.10"
   - "0.12"
+  - "4.0"
+  - "4.1"
 install:
   - npm install
 script:
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Schema validation will be used if you pass an object to any of the validator met
 ```javascript
 req.checkBody({
  'email': {
-  notEmpty: true,
+    notEmpty: true,
     isEmail: {
       errorMessage: 'Invalid Email'
     }
@@ -227,7 +227,7 @@ req.checkBody({
   'password': {
     notEmpty: true,
       isLength: {
-      options: [2, 10] // pass options to the valdatior with the options property as an array
+      options: [2, 10] // pass options to the validator with the options property as an array
     },
     errorMessage: 'Invalid Password' // Error message for the parameter
   },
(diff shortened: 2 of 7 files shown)


ctavan:
Looks good to me!

rustybailey:
Cool, thanks!

andhikamaheva:
`Cannot read property 'sanitizeHeaders' of undefined` when i'm tried using req.sanitizeHeaders('header-name')

gustavohenke:
@andhikamaheva if you started using express-validator since v4.0.0 was released (august 2017), there's a very good chance you're _not_ using what's now called [the Legacy API](https://github.com/ctavan/express-validator/#legacy-api), which is going to be removed soon-ish.

Check [the docs](https://github.com/ctavan/express-validator/) for more info on how to use the newer APIs.

PR #190, opened 2015-12-15
Fixing options not getting passed into chain when using schema

pascalopitz:
call instead of apply does the trick


pascalopitz:
Doh, just saw the test actually wants an options array? Is there a good reason for that?

pascalopitz:
Ok, I see what's going on. I think this should accept both. Will change accordingly, and submit new pul req

PR #191, opened 2015-12-15
Options in schema validation don't have to be an array

pascalopitz:
To facilitate:

    optional: {
        options: { checkFalsy: true }
    },

as an alias of

    optional: {
        options: [{ checkFalsy: true }]
    },

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -265,9 +265,18 @@ function validateSchema(schema, req, loc, options) {
     var paramErrorMessage = schema[param].errorMessage;
     delete schema[param].errorMessage;
 
+    var opts;
+
     for (var methodName in schema[param]) {
       validator.failMsg = schema[param][methodName].errorMessage || paramErrorMessage || 'Invalid param';
-      validator[methodName].call(validator, schema[param][methodName].options);
+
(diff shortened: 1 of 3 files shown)


lib/express_validator.js
```diff
@@ -265,9 +265,18 @@ function validateSchema(schema, req, loc, options) {
...
 
+    var opts;
+
     for (var methodName in schema[param]) {
       validator.failMsg = schema[param][methodName].errorMessage || paramErrorMessage || 'Invalid param';
-      validator[methodName].apply(validator, schema[param][methodName].options);
+
+      opts = schema[param][methodName].options;
+
+      if (opts && !Array.isArray(opts)) {
```
rustybailey: Let's be more explicit here and check for something like `opts != null` instead of just implicitly checking `opts`.

rustybailey:
In the README description for Validation by Schema, I would have a sentence or two explaining how to use the options property properly. Default is have an array of params, but if you only have one param, you don't need to put it in an array.

gustavohenke:
Thank you @pascalopitz and sorry for taking this long to accept!

gustavohenke:
Hi! This is available as of v2.21.0!

PR #202, opened 2016-02-01
Change lodash method names to match 4.x changes

timoruppell:
The _.contains alias was dropped in 4.0.0 in favor of _.includes

This change is compatible with 3.10.x. All the other lodash stuff should be good to go for 4.0.0

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -75,11 +75,11 @@ var expressValidator = function(options) {
 
   // _.set validators and sanitizers as prototype methods on corresponding chains
   _.forEach(validator, function(method, methodName) {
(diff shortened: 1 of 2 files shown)


sahat:
:+1: to this.

rustybailey:
If we're going to do something to conform to a newer version, we might as well upgrade too. Can you push a commit to upgrade lodash to 4.0.x in package.json? If you ping me once the build passes, I'll try to get this in tonight or tomorrow.

timoruppell:
Hi, I quickly went through the additional changes from lodash 4.0 -> 4.2 and you should be good to use 4.2. 

I updated lodash to 4.2.x in package.json and the test pass.

rustybailey:
This is now published as on npm as 2.19.1

PR #208, opened 2016-02-24
Support for defining location of validation for each schema param.

JaniszM:


JaniszM:
Is all fine? Should I do something more to finalise this PR?

README.md
```diff
@@ -246,6 +246,22 @@ req.checkBody({
 });
 ```
 
+You can also define location against to validate specific schema validator by adding `in` parameter as fallowing:
+
+```javascript
+req.checkBody({
```
rustybailey: This example makes the new feature seem confusing and counterintuitive. I would recommend giving you main example with `check()` and then have a note saying that if you happen to use `checkBody/Query/Params`, the `in` attribute will take precedence.

README.md
```diff
@@ -246,6 +246,22 @@ req.checkBody({
 });
 ```
 
+You can also define location against to validate specific schema validator by adding `in` parameter as fallowing:
```
rustybailey: Change to:
You can also define a specific location to validate against in the schema by adding `in` parameter as shown below:

README.md
```diff
@@ -246,6 +246,22 @@ req.checkBody({
...
+    in: 'query',
+    notEmpty: true,
+    isEmail: {
+      errorMessage: 'Invalid Email'
+    }
+  }
+```
+
+This will determine location for the specific validator no matter which method you use to call validator. If you use `in: 'query'` then checkQuery() will be called even if you called you schema as checkParams(). Currently supported location here are `'body', 'params', 'query'`.
+If you provide not supported location parameter then validation process for current parameter will be skipped.
```
rustybailey: Change to:
If you provide a location parameter that is not supported, the validation process for current parameter will be skipped.

lib/express_validator.js
```diff
@@ -259,13 +263,32 @@ var expressValidator = function(options) {
...
+
+    currentLoc = currentLoc === 'any' ? locate(req, param) : currentLoc;
+    var validator = new ValidatorChain(param, null, req, currentLoc, options);
     var paramErrorMessage = schema[param].errorMessage;
     delete schema[param].errorMessage;
 
     for (var methodName in schema[param]) {
+      // skip method if this is location definition
+      if (methodName === 'in') {
+        currentLoc = loc;
```
rustybailey: Maybe I'm missing something, but why are we redefining `currentLoc` here? Shouldn't it be finalized on line 274?

rustybailey:
You have tests that check for using `in` within the schema for `check()`. I would also have a test that checks if `in` works within a specific check like `checkBody`.

lib/express_validator.js
JaniszM: In line 274 we are defining location per one parameter in whole schema, and this is only in case when `in` is defined. So for rest parameters we should use old, original location e.g.: user is calling `req.checkBody(schema);` but one of the parameters in schema has `in: 'query'`. Without reassigning original location after reading param with `in` rest of params (this without `in`), will be checked against query as well, but should be checked against body :)

JaniszM:
Ok, doc fixed, tests added. I added simple tests to check if other check() method works in a specific location. Should be enough :)

rustybailey:
Can you please merge in latest master?

JaniszM:
Here you go :)

Issue #203, opened 2016-02-14
'WithMessage'  not work for custom asynchronous validators.


Maxim-Chugaev:
Asynchronous validator
```
Channel.channelNameAvailable = (name, id = 0)->
  Channel
    .query({'where':{'name': name}, 'whereNot': {'id': id}})
    .fetch()
    .then (channel)->
      Promise.reject('channel name already in use') if channel?
```
Initialize express validator
```
router.use expressValidator
                  customValidators:
                    channelNameAvailable: Channel.channelNameAvailable
```
Validation

```
req
        .check('name')
        .notEmpty().withMessage('channel name is required')
        .isLength({min: 4, max: 20}).withMessage('channel name must have length in range between 4 and 20')
        .channelNameAvailable(req.body.id).withMessage('channel name already in use')
```

if channel name is already exists, i get error 'Invalid value', but must be another message!

LordWingZero:
I am also having this issue. Came here to post and saw this.

Please fix.

rustybailey:
@Jakeii or @chrissinclair -- Would one of you guys be able to help with this?

chrissinclair:
I'll see if I can't take a look over the weekend. @Maxim-Chugaev @LordWingZero, do either of you have a minimal example showing this already I can use as a base?

LordWingZero:
# Bare-bones test case showcasing error:

## app.js
```Javascript
var express = require('express');
var bodyParser = require('body-parser');
var expressValidator = require('express-validator');
var Promise = require("bluebird");

var app = express();

app.use(bodyParser.json());
app.use(bodyParser.urlencoded({
    extended: true
}));
app.use(expressValidator({
    customValidators: {
        testValidator: function(username) {
            return new Promise(function(resolve, reject) {
                reject();
            });
        }
    }
}));

app.get('/', function (req, res) {
    res.send([
        '<form method="post">',
        '<input name="testInput" />',
        '</form>',
        ].join(''));
});

app.post('/', function(req, res) {
    
    req.checkBody('testInput', 'Default Message')
        .testValidator()
        .withMessage('With Message');

    req.asyncValidationErrors()
    .then(function(){
        return res.send('Validation succeeded.');        
    })
    .catch(function(errors) {
        return res.send(errors);
    });

});

app.listen(process.env.PORT, process.env.IP, function() {
    console.log('Example app listening on port 3000!');
});
```

## package.json
```JSON
{
  "name": "express-validator-test-case",
[...]

PR #212, opened 2016-03-05
#203 - Fixed bug where .withMessage() would not apply to a custom async validator

chrissinclair:
The idea here is to use the async parallel of how `withMessage()` works in the synchronous case. We expand `lastError` to include a flag indicating whether the validator is async, and if so pop the promise from `_asyncValidationErrors` and replace it with one that resolves to the provided message. 

As we have already attached a handler to the promise coming from the custom validator, we have to route the existing one to an empty catch handler, otherwise bluebird will throw an 'unhandled rejection' error.


chrissinclair:
Oh - only just saw this. Will take a look tonight.

chrissinclair:
That's better! @rustybailey - are you able to take a look?

rustybailey:
Yes, should be able to look at this tonight or tomorrow morning. Thank you so much for doing this!

PR #215, opened 2016-03-13
Upgraded bluebird

AuspeXeu:
If I understood the functionality of reflect correctly, this should be the fix needed for the upgrade. It's a bit unfortunate that they took down the documentation for `.settle()` so that I cannot really look the semantics up.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -174,6 +174,8 @@ var expressValidator = function(options) {
     req.asyncValidationErrors = function(mapped) {
       return new Promise(function(resolve, reject) {
         var promises = req._asyncValidationErrors;
+        // Migrated using the recommended fix from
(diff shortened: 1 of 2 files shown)


lib/express_validator.js
```diff
@@ -174,7 +174,9 @@ var expressValidator = function(options) {
     req.asyncValidationErrors = function(mapped) {
       return new Promise(function(resolve, reject) {
         var promises = req._asyncValidationErrors;
-        Promise.settle(promises).then(function(results) {
+        Promise.all(promises.map(function(promise) {
```
rustybailey: Since we had to move from using a well-documented method to now using their recommended fix, can you add a comment above this line just explaining what this chunk of code does?

AuspeXeu: Sure thing!

PR #226, opened 2016-04-08
Fixing warning from node-validator

lobodart:
If you try to check an `undefined` field (the goal of the library in a nutshell), you get this warning from the `node-validator` library :

> validator deprecated you tried to validate a undefined but this library (validator.js) validates strings only. Please update your code as this will be an error soon. node_modules/express-validator/lib/express_validator.js:338:41

Code sample :

```javascript
req.assert('firstName', "missing_firstName").notEmpty(); // The error comes from here if this param is undefined
req.assert('lastName', "missing_lastName").notEmpty();
req.assert('username', "missing_username").notEmpty();
req.assert('password', "missing_password").notEmpty();

req.assert('username', "invalid_username").isEmail();
req.assert('password', "password_too_short").len(6);
```

This commit should fix the problem.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -335,7 +335,7 @@ function makeValidator(methodName, container) {
     args.push(this.value);
     args = args.concat(Array.prototype.slice.call(arguments));
 
-    var isValid = container[methodName].apply(container, args);
+    var isValid = !args[0] ? false : container[methodName].apply(container, args);
     var error = formatErrors.call(this, this.param, this.failMsg || 'Invalid value', this.value);
 
     if (isValid.then) {


rustybailey:
This is close. However, if you check one of the builds (https://travis-ci.org/ctavan/express-validator/jobs/121801347), you see we still have warnings regarding numbers:
```
validator deprecated you tried to validate a number but this library (validator.js) validates strings only. Please update your code as this will be an error soon. lib/express_validator.js:382:40
```

Since the main issue is that validator.js doesn't automatically convert to strings anymore in v5 (see https://github.com/chriso/validator.js/pull/496), we need to find a good place in the code to convert the params to strings, so that validator won't shout at us. However, when doing this we need to make sure that `optional()` still works as intended. Once we fix this, we can also upgrade validator to v5.x.

fvargas:
As was mentioned above, the problem is that validator.js is now designed to work only with strings. However, express-validator happily passes along non-string data to the validators, which is why we see that warning.

In the case of `undefined` values, the first thing to note is that despite seeing that warning, the validators are still being run. The `undefined` value is being coerced to an empty string, and the validators are running with that empty string as their input. This may or may not be what one expects or wants.

Second, what your code is doing is labeling all falsy values as errors. So I expect that if you test your code with `""`, `null`, `0`, or any other falsy value, it will produce an error even when it is not appropriate. Consider the input `0`; this should actually be treated as the string `"0"` and be validated accordingly.

Finally, I agree the way `undefined` values are currently handled is problematic. My concerns and solution are explained in #227.

uptownhr:
Any update here? 

Also what is the expected validation error if you pass in a `undefined` object?

hubsn:
+1

I'm looking forward to disappear this annoying warning message.
@rustybailey Could you tell me when fix this?

rustybailey:
Sorry, I've been super busy and haven't had time for code outside of work. I'll try to look at this tonight.

theHarvester:
+1 Still getting warnings. In my case, the input is coming from JSON. I can't control if the JSON request has numbers or nulls in it.

rustybailey:
The warnings have been fixed in 2.20.4.

Issue #227, opened 2016-04-13
No method for validating existence of property in req.params, req.body, or req.query


fvgs:
Suppose I have a route expecting to receive the parameter `req.body.heading`. If this parameter exists and was sent by the client, then my validation will run as expected. I will check `req.validationErrors()` and proceed accordingly. However, if `req.body.heading` was not sent by the client, its value will be `undefined`. Based on my limited testing, when this happens, the `undefined` value will be coerced to an empty string and validation will proceed on that empty string. A `validator deprecated` message also gets printed to the console, and I understand this is going to be addressed in the future. However, the problem here is that there is currently no way to check if a value is `undefined` and produce an error when it is `undefined`. Currently, the `undefined` value gets coerced to an empty string and validated as an empty string.

This is problematic because there is a big difference between not receiving a value from the client and receiving an empty string. These constitute independent cases that one may wish to handle differently. In my case, I consider receiving no value (i.e. `undefined`) as an error, whereas receiving an empty string may or may not be valid depending on the context.

My solution has been to create a custom validator to check if the value exists:

``` javascript
customValidators: {
  exists: value => value !== undefined,
}
```

This achieves the desired behavior whereby if the value is `undefined`, an error will be produced in `req.validationErrors()`. In practice, I also use the `optional()` method immediately after `exists()` to stop any further validation when the value is `undefined`.

Typical usage would look something like this:

``` javascript
req.checkBody('heading').exists().optional()...;
```

With that in mind, I'd appreciate some feedback on whether or not this is the best way to address this issue. Likewise, if this makes sense and doesn't conflict with any upgrade plans, I'd like to make a PR to add an `exists()` method. In this PR I would go ahead and bundle the `optional()` behavior into the `exists()` method since using it without that behavior doesn't really make sense. Though I would add an option to allow `exists()` to be used without the `optional()` behavior.

srkimir:
@fvgs thank you on very nice explanation of the problem. Can you please confirm do you still have described problem. As you said: 
> However, the problem here is that there is currently no way to check if a value is undefined and produce an error when it is undefined

So lets say that you want to check `email` field was sent by client, so your validation may look like
```
req.checkBody('email', 'valid email address is required').isEmail()
```
And lets say that request body is empty `{}` so email value is `undefined`, this will produce error like
```
{
  "email": {
    "param": "email",
    "msg": "valid email address is required"
  }
}
```
So it will produce an error if value is `undefined`

fvgs:
[...]

PR #229, opened 2016-04-30
Upgrade node-validator to v5

rustybailey:
Upgrade node-validator to v5 and automatically coerce values to strings on our end since node-validator doesn't do it anymore, but requires strings as input.

Would love a thorough review by a few of these people if possible: @ctavan, @fvargas, @Jakeii, @chrissinclair, @AuspeXeu, @JaniszM 

Did I miss anything? Should this be a major or minor version bump?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 An [express.js]( https://github.com/visionmedia/express ) middleware for
 [node-validator]( https://github.com/chriso/validator.js ).
 
+*Note: As of node-validator's v5, it does not automatically coerce values to strings, but it still requires strings as input. So now express-validator will automatically coerce your input to strings before to passes to node-validator. Ideally, `notEmpty()` should still function as normal.*
+
 ## Installation
 
 ```
@@ -119,18 +121,18 @@ Define your custom validators:
 ```javascript
 app.use(expressValidator({
  customValidators: {
-    isArray: function(value) {
-        return Array.isArray(value);
+    saysHello: function(value) {
+        return value === "Hello";
     },
     gte: function(param, num) {
-        return param >= num;
+        return Number(param) >= Number(num);
     }
  }
 }));
 ```
 Use them with their validator name:
 ```javascript
-req.checkBody('users', 'Users must be an array').isArray();
+req.checkBody('users', 'Users must be an array').saysHello();
 req.checkQuery('time', 'Time must be an integer great than or equal to 5').isInt().gte(5)
 ```
 ####`customSanitizers`
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -85,6 +85,13 @@ var expressValidator = function(options) {
   });
 
   ValidatorChain.prototype.notEmpty = function() {
+    if (this.value == null) {
+      var error = formatErrors.call(this, this.param, this.failMsg || 'Invalid value', this.value);
+      this.validationErrors.push(error);
+      this.req._validationErrors.push(error);
+      this.lastError = { param: this.param, value: this.value, isAsync: false };
+      return this;
+    }
     return this.isLength({
       min: 1
     });
@@ -332,7 +339,7 @@ function makeValidator(methodName, container) {
     }
 
     var args = [];
-    args.push(this.value);
+    args.push(this.value += '');
     args = args.concat(Array.prototype.slice.call(arguments));
 
     var isValid = container[methodName].apply(container, args);
@@ -377,7 +384,7 @@ function makeSanitizer(methodName, container) {
     var result;
     this.values.forEach(function(value, i) {
       if (value != null) {
-        var args = [value];
+        var args = [value += ''];
         args = args.concat(Array.prototype.slice.call(_arguments));
         result = container[methodName].apply(container, args);
 
@@ -433,7 +440,7 @@ function formatParamOutput(param) {
   if (Array.isArray(param)) {
     param = param.reduce(function(prev, curr) {
       var part = '';
-      if (validator.isInt(curr)) {
+      if (validator.isInt(curr.toString())) {
         part = '[' + curr + ']';
       } else {
         if (prev) {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "bluebird": "3.3.x",
     "lodash": "4.6.x",
-    "validator": "4.9.x"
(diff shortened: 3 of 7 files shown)


chrissinclair:
I think this covers the scenarios and should cover off the change to the validator library. The only issue I can see (and I'm guessing you've seen it to) is checking more complex objects - e.g. arrays. The only way I could get an isArray custom validator to work was 

```javascript
isArray: function(value) {
  return Array.isArray(JSON.parse('[' + value + ']'));
},
```

You'll face similar issues if you were expecting to validate objects (e.g. from a parsed JSON in the body of a request): 

```javascript
test: function(value) {
  return value.foo > value.bar;
}
```

will now break as rather than value being something like `{foo: 10, bar: 12}`, instead it will be `'[object Object]'`. I can't see a way to recover this, other than to re-write the validator entirely to not need the object (which although we could presumably do in this case, if you had something more complex I'm not sure you always could). 

Given that this would break existing validators, I think this needs to be a major version bump.

Jakeii:
Can we not just re add the stuff they removed from node-validatior?

This will make it work as before (the current tests pass at least)

```
var validator = require('validator');
var _ = require('lodash');
var Promise = require('bluebird');

validator.extend = function(name, fn) {
  validator[name] = function() {
    var args = Array.prototype.slice.call(arguments);
    args[0] = validator.toString(args[0]);
    return fn.apply(validator, args);
  };
};

validator.init = function() {
  for (var name in validator) {
  if (typeof validator[name] !== 'function' || name === 'toString' ||
    name === 'toDate' || name === 'extend' || name === 'init' ||
    name === 'isServerSide') {
     continue;
  }
  validator.extend(name, validator[name]);
  }
};

validator.toString = function(input) {
   if (typeof input === 'object' && input !== null && input.toString) {
    input = input.toString();
   } else if (input === null || typeof input === 'undefined' || (isNaN(input) && !input.length)) {
    input = '';
   }
   return '' + input;
 };

 validator.toDate = function(date) {
  if (Object.prototype.toString.call(date) === '[object Date]') {
    return date;
  }
  date = Date.parse(date);
  return !isNaN(date) ? new Date(date) : null;
 };

validator.init();
```

README.md
```diff
@@ -5,6 +5,8 @@
 An [express.js]( https://github.com/visionmedia/express ) middleware for
 [node-validator]( https://github.com/chriso/validator.js ).
 
+*Note: As of node-validator's v5, it does not automatically coerce values to strings, but it still requires strings as input. So now express-validator will automatically coerce your input to strings before to passes to node-validator. Ideally, `notEmpty()` should still function as normal.*
```
fvargas: This note may be unnecessary. The behavior (after some adjustments) should be just as it was previously, only that with the upgrade of [validator.js](https://github.com/chriso/validator.js) (note the new name), the string coercion is performed by express-validator rather than validator.js. The user need not be concerned with this detail since the interface has not changed.

lib/express_validator.js
```diff
@@ -85,6 +85,13 @@ var expressValidator = function(options) {
   });
 
   ValidatorChain.prototype.notEmpty = function() {
+    if (this.value == null) {
```
fvargas: For undefined or null values, validator.js would coerce them to empty strings and perform validation on them as empty strings. This code changes that behavior by making null and undefined values throw an error instead. This would be a major change.

For this PR, I don't think major changes should be made so users can safely upgrade to this version and get rid of that annoying warning without having to modify their code. I'll discuss this further in a PR comment as well as my thoughts on how best to address undefined values for a later PR.

lib/express_validator.js
```diff
@@ -332,7 +339,7 @@ function makeValidator(methodName, container) {
     }
 
     var args = [];
-    args.push(this.value);
+    args.push(this.value += '');
```
fvargas: It may make more sense to perform the conversion and assignment of `this.value` on [line 23](https://github.com/ctavan/express-validator/pull/229/files#diff-6c47d849cc78d146191d4d361699b452L23) so it need only be converted once at the source.

If that is not an option, I would suggest moving the assignment onto its own line in order to make it more explicit.
e.g.
``` javascript
this.value += '';
args.push(this.value);
```

On a more important note, this means of converting the value to a string is insufficient and introduces breaking changes. The behavior changes for values such as undefined, null, and NaN, and possibly for arrays and objects as others have noted. For instance, validator.js would coerce undefined, null, and NaN values to an empty string. Whereas this code coerces undefined to the string `'undefined'`.

As I mentioned above, I think it's important to preserve the same conversion rules for this PR so as to not introduce major changes. Therefore, I think @Jakeii's suggestion to mirror the code previously used by validator.js makes sense.

lib/express_validator.js
```diff
@@ -377,7 +384,7 @@ function makeSanitizer(methodName, container) {
     var result;
     this.values.forEach(function(value, i) {
       if (value != null) {
-        var args = [value];
+        var args = [value += ''];
```
fvargas: Similar to the above, it may make more sense to perform the conversion on [line 43](https://github.com/ctavan/express-validator/pull/229/files#diff-6c47d849cc78d146191d4d361699b452L43). Or better yet, DRY and just perform the conversion once for both situations.

Moreover, the assignment here seems unnecessary since the `value` argument is not used further. Though if the desire was to create an assignment for the sake of maintainability, I would do so in its own line, using a new variable so as to not mutate the argument:
``` javascript
strValue += '';
```

Though the first solution is still preferable.

fvargas:
I've outlined the issues I noticed above. Though I'll say that, not being too familiar with the codebase, I was not checking very heavily for potential omissions.

Anyway, my recommendation would be to make this a minor version change so users can get rid of the annoying warning from validator.js without having to modify their existing code. This basically means doing as @Jakeii suggested and mirroring the conversion mechanisms previously used by validator.js.

Additionally, in the `notEmpty()` function, @rustybailey partially addresses the point I brought up in #227 about treating undefined values as errors rather than allowing them to be coerced to an empty string. I do believe this should be implemented (not necessarily as part of `notEmpty()`), but in a separate PR so as to not introduce major changes to this one. With some feedback, I'd be happy to work on that in another PR.

rustybailey:
To everyone invested in this: I opened a new PR with @Jakeii's suggestion. Please take a look over there with comments or thumbs up: https://github.com/ctavan/express-validator/pull/231/files

PR #231, opened 2016-05-05
Update node-validator to v5 and restore automatic string coercion.

rustybailey:
Added lines from @Jakeii's suggestion in https://github.com/ctavan/express-validator/pull/229, so that all tests pass with an upgrade to node-validator v5.


chrissinclair:
Makes sense - and this way I don't think it needs to be a major version bump :)

PR #232, opened 2016-05-20
support for cookies

parisholley:


rustybailey:
1. We need tests for `checkCookies()`.
2. Are we assuming that anyone that wants to check/sanitize cookies is using cookie-parser? Just want to make sure there aren't any other cookie parsers we're not thinking about or if cookie-parser is the de facto lib.

gustavohenke:
> [...] or if cookie-parser is the de facto lib

It seems like it is; the [req.cookies docs](http://expressjs.com/en/4x/api.html#req.cookies) mention about it, so I imagine most users will just install it when they need cookies support.

Anyway, we will still need tests for `checkCookies`, can you write them @parisholley?

PR #237, opened 2016-06-08
Update production depedencies

KevinBusse:
The company I work for doesn't like packages with outdated dependencies, but I'd hate to loose this package.


rustybailey:
Updated in v2.20.6

KevinBusse:
Thanks

Issue #242, opened 2016-06-18
paramErrorMessage will be deleted permanently from schema after first validation if schema is not recreated on every new request


srkimir:
In scenario where schema is not recreated on every request `paramErrorMessage` will be permanently deleted from schema. To me it seems perfectly valid scenario to use predefined schemas instead of recreating schemas over and over again on each request. In following example only first ever request will have expected param error message `"msg": "a must be integer greater or equal to 1",` all following requests will return `"msg": "Invalid param"`. This is caused by line number: `343` in https://github.com/ctavan/express-validator/blob/master/lib/express_validator.js. Please check end of answer for proposal how this can easily be solved. **Example:**

_// schemas/index.js_
```
'use strict'

exports.somethingSchema = {
  a: {
    isInt: {
      options: [{ min: 1}]
    },
    errorMessage: 'a must be integer greater or equal to 1'
  }
}
```

_// test.js_
```
const test = (req, res, next) => {
  req.checkBody(schemas.somethingSchema)
  const errors = req.validationErrors()
  
  if (errors) {
    return res.status(422).json({
      type: 'validation-error',
      message: 'Constraint violation',
      errors
    })
  }
  return next()
}

app.post('/something', test, (req, res, next) => {
  res.send('ok')
})
```

_How it can be solved:_ Clearly intention with using `delete` is that `errorMessage` not ends like `methodName` in `for (var methodName in schema[param]) {...}`. Instead `delete` this can be solved in more safer way with `if (methodName === 'in' || methodName === 'errorMessage') { ... continue; }`
One more reason why i would vote against `delete` is that you never know how schema will be used after validation, what if application that is using your library later in request is using `paramErrorMessage`...

rustybailey:
Should be fixed in 2.20.8

PR #243, opened 2016-06-18
avoid deletion of param error message when using schema validation

srkimir:
Please check open issue https://github.com/ctavan/express-validator/issues/242 for example and details.
Changes from PR will not `delete` param error message from schema, it will skip validation if `methodName` is equal to `errorMessage`. On this way same predefined schema in application can be reused safely.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -340,7 +340,6 @@ function validateSchema(schema, req, loc, options) {
     currentLoc = currentLoc === 'any' ? locate(req, param) : currentLoc;
     var validator = new ValidatorChain(param, null, req, currentLoc, options);
     var paramErrorMessage = schema[param].errorMessage;
-    delete schema[param].errorMessage;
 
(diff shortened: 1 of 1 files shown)


rustybailey:
Looks like the delete was added here: https://github.com/ctavan/express-validator/commit/8ff36987ce13476fad5dbb5653612860ecf0aca3#diff-6c47d849cc78d146191d4d361699b452

@Jakeii What was the reason that was added?

Jakeii:
@rustybaily It was to prevent validation of the `errorMessage` property, I'm not sure why I went with delete. Later on someone added the prevention for `in` validation, with a better approach.

srkimir:
@rustybailey @Jakeii thanks on your response
Instead `delete` PR use same logic to prevent validation for `errorMessage` like someone else did for `in`. Maybe we could merge this

rustybailey:
:+1:

PR #244, opened 2016-06-20
Fix a bug for body schema checking

lovanwubing:
When my schema has options, the options won't be passed to my custom validators. By looking through the code and found out the wrong usage of apply and call.

Please merge it to master. thanks


rustybailey:
Are you passing the options as an array? That is how schema options are used in the README.

lovanwubing:
OK. Thanks, I got it!

PR #248, opened 2016-07-02
Fix typo

felipelrz:


rustybailey:
I think this is the intended order. No typo here.

PR #250, opened 2016-07-24
Upgraded validator to 5.5.x

Mosnar:
Hey, remember when stackoverflow went down the other day? It's because of a bug in trimming strings, which is fixed in validator 5.5. Read more: http://stackstatus.net/post/147710624694/outage-postmortem-july-20-2016


rustybailey:
Upgraded further in https://github.com/ctavan/express-validator/pull/256

PR #253, opened 2016-08-02
String formatting for error messages

dpolivy:
In prior versions of express-validator, you could include tokens like
%0, %1, etc in error message, and validator.js would replace them with
the value of the input or parameters passed to the validators.

This commit restores that behavior.

%0 is the value of the parameter being tested
%1... are the values of the parameters passed to the validator function
itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,23 @@ mappedErrors:
 ```
 *Note: Using mappedErrors will only provide the last error per param in the chain of validation errors.*
 
+### String formatting for error messages
+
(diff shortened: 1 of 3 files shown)


rustybailey:
This is awesome. Can you add documentation to the README for this feature?

dpolivy:
@rustybailey Done. Let me know if that's sufficient.

rustybailey:
Appreciate it. This is now in 2.20.10.

PR #255, opened 2016-08-26
Nesting check that requires all children to fail

michaeldfallen:
This pull request proposes a way of overriding a set of child checks with a parent check if all the children fail.

The proposed interface to this is a `ValidatorChain` named `all`, which accepts a list of `ValidatorChains` and checks that they all failed. If they all did it removes them from the requests `_validationErrors` list and adds it's own error in their place.

Example:

```
req.check('name', 'Please enter your name').all([
    req.check('firstName', 'Enter your first name').notEmpty(),
    req.check('lastName', 'Enter your last name').notEmpty()
]);
```

---

**Background**

This comes from the approach to errors that I've previously used on GDS services where we want to provide error messages that are as human readable as possible. For example:
- Providing no input at all gives: _Please enter your date of birth_

image
- Only filling in some of the fields gives errors for each field

image

This is a simple example but they can get quite complex. The aim is always to allow the message to the user to be as simple and understandable as possible.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -182,6 +182,26 @@ var expressValidator = function(options) {
     return this;
   };
 
+  ValidatorChain.prototype.all = function(childChains) {
+    var childrenFailed = _.every(childChains, function(chain) {
+      return chain.validationErrors.length > 0;
+    });
+    if (childrenFailed) {
+      var parentError = formatErrors.call(this, this.param, this.failMsg, this.value);
+      var failed = _.map(childChains, function(chain) {
+        return chain.param;
+      });
+      var cleanedErrors = this.req._validationErrors.filter(
+        function(error) {
+          return !_.includes(failed, error.param);
+        }
+      );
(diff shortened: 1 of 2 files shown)


michaeldfallen:
Uncertain who's lead maintainer so poking owner and most active: @ctavan @rustybailey.

rustybailey:
Sorry, been super busy. Thanks for adding tests, but since this is a new feature, can you also add documentation to the README?

rustybailey:
Also, can you fix the style issues that are causing your build to fail?

gustavohenke:
@michaeldfallen, do you care to finish this PR by adding what Rusty commented?

michaeldfallen:
Yes, will do. Sorry, been busy here myself and we ended up using a different approach for the checks we wanted this feature to help with. 

As long as you find it useful (since I won't be using it day to day) I'll finish this up and get it clean.

Closing for now until I get time to fix this.

Issue #259, opened 2016-09-21
req.check(schema) does not reliably catch failed validations


helmutkian:
When using `req.check` instead of `req.checkBody`, the library fails to catch certain validation errors.

````
req.check({
   "type":{
      "optional":true,
      "isIn":{
         "options":[
            [
               "A",
               "B"
            ]
         ]
      },
      "errorMessage":"ERROR"
   },
   "name":{
      "optional":true,
      "isLength":{
         "options":[
            {
               "min":0,
               "max":100
            }
         ]
      },
      "errorMessage":"ERROR"
   },
   "id":{
      "optional":true,
      "isUUID":{
         "options":[
            4
         ]
      },
      "errorMessage":"ERROR"
   },
   "user_id":{
      "optional":true,
      "isUUID":{
         "options":[
            4
         ]
      },
      "errorMessage":"ERROR"
   },
   "group_id":{
      "optional":true,
      "isUUID":{
         "options":[
            4
         ]
      },
      "errorMessage":"ERROR"
   }
});
````

Sending a POST with `{ "group_id": "INVALID" }` yields no errors. However `{ "type": "C", "group_id": "INVALID" }` will yield the two correct errors. However swapping `req.check` with `req.checkBody` seems to solve the issue.

PR #260, opened 2016-09-22
Fixes optional params breaking req.check

helmutkian:
Fixes issue with: https://github.com/ctavan/express-validator/issues/259

BUG: If a schema with `optional: true` is passed to `req.check` and the optional parameter is omitted from the request, then the `locate` function will return `undefined` for that parameter and all subsequent validations will not run.


helmutkian:
Done :)

rustybailey:
Thanks again. This is now in 2.20.10

PR #261, opened 2016-09-30
add sanitize option in validate schema

joel-leandev:
Added sanitize option in validate schema that will sanitize the param.

var Schema ={
         'duration':{
            'optional' : true,
            'sanitize' : ['escape']
        }
};

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -361,7 +361,21 @@ function validateSchema(schema, req, loc, options) {
       }
 
       validator.failMsg = schema[param][methodName].errorMessage || paramErrorMessage || 'Invalid param';
-      validator[methodName].apply(validator, schema[param][methodName].options);
+
+      if(validator[methodName]){
+        validator[methodName].apply(validator, schema[param][methodName].options);
+      }
+      else if(methodName==='sanitize'){
+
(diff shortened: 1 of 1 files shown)


gustavohenke:
- Don't use `eval`
- Write some tests for this feature

Also please check the Travis build logs for your PR to see what went wrong and fix it.

lib/express_validator.js
```diff
@@ -361,7 +361,21 @@ function validateSchema(schema, req, loc, options) {
...
+
+      if(validator[methodName]){
+        validator[methodName].apply(validator, schema[param][methodName].options);
+      }
+      else if(methodName==='sanitize'){
+
+        for(var index in schema[param][methodName]){
+          var currentSanitize = schema[param][methodName][index];
+          var funcName = "validator.req.sanitize('"+param+"')."+currentSanitize+"()";
+          eval(funcName);
```
gustavohenke: Never use `eval()`.

gustavohenke:
I'm not going to merge this one.
With v4 release, the legacy API is being deprecated, and sanitization is not having lots of space in the new APIs (since most of the time the amount of code makes it better if you sanitize data directly).

PR #265, opened 2016-10-14
Fixed error in README (async validation)

joehull:
Fixed error in the Asyncronous validation example. The promise was resolving in the case of user existing which should do the opposite

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ app.use(expressValidator({
       return new Promise(function(resolve, reject) {
         User.findOne({ username: username })
         .then(function(user) {
-          if (user) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,7 +203,7 @@ app.use(expressValidator({
       return new Promise(function(resolve, reject) {
         User.findOne({ username: username })
         .then(function(user) {
-          if (user) {
+          if (!user) {
             resolve(user);
```
gustavohenke: What do you think of even removing the `user` parameter from this `resolve()` call?
It doesn't make much sense to resolve something which is probably undefined :)

Issue #249, opened 2016-07-07
Provide a cleaner async validation implementation


cbovis:
The current implementation of async validation results in some pretty messy code. 

Throwing an exception to handle validation errors feels wrong since they're not exceptions, they're expectations. More on this here... http://stackoverflow.com/questions/729379/why-not-use-exceptions-as-regular-flow-of-control

An example of current implementation,

```
Place.findById(req.params.placeId).then(place => {
  return req.asyncValidationErrors().then(() => {
    res.render('places/edit', {
      place
    });
  }).catch(err => {
    if (Array.isArray(err)){
      res.render('places/edit', {
        place,
        errors: err
      });
    } else {
      throw err;
    }
  });
}).catch(err => { next(err); });
```

Suggested implementation,

```
Promise.join(
  Place.findById(req.params.placeId),
  req.validate(),
  (place, errors) => {
    res.render('places/edit', {
      place,
      errors
    });
  }
).catch(err => { next(err); });
```

Using yield syntax via co-express the second method would produce even cleaner code as opposed to the messy try-catch solution that is presently required,

```
var place = yield Place.findById(req.params.placeId);
var errors = yield req.validate();

res.render('places/edit', {
  place,
  errors
});
```

By passing the errors back via the promise resolution you can see the code becomes much cleaner. If you're interested in implementing this then I'm happy to put together a pull request for submission.

Thoughts?

gustavohenke:
I don't see your idea only as a way to make the code cleaner, but also more consistent.
When you do sync validation, express-validator will not throw exceptions. So why throw them when going async?

gustavohenke:
Additionally, as per [semver](http://semver.org/), this would require a major version bump.

cbovis:
I have some spare time right now whilst looking for work so I'm going to fork and take a look at implementing this 👍

cbovis:
Implemented via pull request #269

gustavohenke:
Closed by #269 😉

PR #269, opened 2016-11-02
Introducing friendlier API for retrieving validation errors

cbovis:
As mentioned in issue #249 the API for retrieving validation errors is rather cumbersome. 

To solve this problem I've introduced a new `req.getValidationErrors()` method which is async by default and returns errors as part of the promise resolution. This makes it easier to write readable code, especially when coupled with generators, and keeps code consistent when switching between sync/async.

For example, using generators with the new method code is now written as such:

```javascript
var errors = yield req.getValidationErrors();

if (errors) {
  res.send('There were errors');
} else {
  res.send('All is well');
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,9 +341,10 @@ req.assert('email', 'Invalid email')
     .notEmpty().withMessage('Email is required')
     .isEmail();
 
-req.getValidationErrors(function(errors){
-  // do something with errors
-});
+req.getValidationErrors()
+   .then(function(errors){
+     // do something with errors
+   });
 
 ```
 errors:
--- a/README.md
+++ b/README.md
@@ -48,15 +48,18 @@ app.post('/:urlparam', function(req, res) {
   // OR find the relevent param in all areas
   req.sanitize('postparam').toBoolean();
 
-  var errors = req.validationErrors();
-  if (errors) {
-    res.send('There have been validation errors: ' + util.inspect(errors), 400);
-    return;
-  }
-  res.json({
-    urlparam: req.params.urlparam,
-    getparam: req.params.getparam,
-    postparam: req.params.postparam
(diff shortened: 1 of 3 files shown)


gustavohenke:
Wow, amazing work!

I will later review it with more attention.
At a first glance, I think this could be the start of v3.0.0.

README.md
```diff
@@ -189,49 +192,6 @@ Same as [req.check()](#reqcheck), but only looks in `req.params`.
...
-
-req.asyncValidationErrors()
-.then(function() {
-// create user
-})
-.catch(function(errors) {
-  res.send(errors);
-});
-
-```
```
gustavohenke: I wouldn't remove this full block. Perhaps leave the part which demonstrates how to create a custom async validator.

README.md
```diff
@@ -380,7 +340,11 @@ You can provide an error message for a single validation with `.withMessage()`.
 req.assert('email', 'Invalid email')
     .notEmpty().withMessage('Email is required')
     .isEmail();
-var errors = req.validationErrors();
+
+req.getValidationErrors(function(errors){
```
gustavohenke: Your method doesn't accept a callback, so please remove this example or convert it to promise style

test/getValidationErrorsTest.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// This before() is required in each set of tests in
+// order to use a new validation function in each file
+before(function() {
+  delete require.cache[require.resolve('./helpers/app')];
+  app = require('./helpers/app')(validation);
+});
+
+describe('#validateTest()', function() {
```
gustavohenke: Please rename this suite in order to match the implemented method name.

PR #282, opened 2016-11-21
Bug Fix: Optional validate method may not be applied when using a schema

### LINKED ISSUE #222: Optional validate method may not be applied when using a schema.
The validateSchema function iterates over a param object's keys when using a schema. If you have rules that come before optional they will be used before it can be determined that validation should be skipped because you specified optional in your schema.

This will only work if you make sure to specify optional first in your schema object literal, but because JavaScript does not guarantee object property order, this only happens to work because of the behavior of V8 where properties tend to come out in the same order they were put in.

I think it might make sense to check for optional before iterating through a param config's properties.

IOAyman:
As suggested by @srkimir, this PR should fix #222.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -374,6 +374,14 @@ function validateSchema(schema, req, loc, options) {
 
     var opts;
 
+    if (schema[param].optional) {
+      validator.optional.apply(validator, schema[param].optional.options);
+
+      if (validator.skipValidating) {
+        continue; // continue with the next param in schema
+      }
+    }
+
     for (var methodName in schema[param]) {
       if (methodName === 'in') {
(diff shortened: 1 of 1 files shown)


lib/express_validator.js
```diff
@@ -385,9 +393,8 @@ function validateSchema(schema, req, loc, options) {
       }
 
       if (methodName === 'errorMessage') {
-        /**
-         * Also do not validate if methodName
-         * represent parameter error mesage
+        /* Also do not validate if methodName
+         * represent parameter error message
```
gustavohenke: what's the change in this block?

IOAyman: 1. Typo: `mesage` => `message`
2. `/**`  => `/*` (

PR #285, opened 2016-11-24
#getValidationResult() validation result object

### LINKED ISSUE #30: Chaining check causes duplicated errors
`request.check('email', 'Email is not valid').notEmpty().notNull().isEmail();`

With no "email" field set this will return 3 repeatitions of the same message. 
Shouldn't the error message be unified on chained call like this?

Thank you

### LINKED ISSUE #223: Warning: a promise was rejected with a non-error: [object Array]
Working with native Promise i got this warning.

`Warning: a promise was rejected with a non-error: [object Array]`

If validation errors exists, `onReject` function receive an array of this errors. https://github.com/ctavan/express-validator/blob/master/lib/express_validator.js#L155

Can we define own "class" of error? ex. ExpressValidationErrors which will contain property ._list (array) of errors. And .all() will return this list.

``` js
req.asyncValidationErrors().catch(errors => {
  req.flash('errors', errors.all());
  renderEdit(res);
});
```

### LINKED ISSUE #235: how to show only one validation error message per field
I have a registration form and a user has to submit email address in the form (along with other fields). my validation rule for the email field with 'express-validator' is

`req.checkBody('email', 'Email field is required').notEmpty().isEmail().withMessage('Invalid email address');`

and I am using jade to print out the errors as (with bootstrap)

```
  if errors.length  
    .alert.alert-danger
      a.close(href='#', data-dismiss='alert', aria-label='close') ×
      for error in errors
         div=error.msg 
```

the above code print out all the form errors of the fields, now the problem is, if email field is empty, it shows two error messages 'Email field is required' and 'Invalid email address'

I want to evaluate the second validation isEmail() only if the email is not empty. If empty then it need to show only one error message 'Email field is required'. Is there any solution to this ?

gustavohenke:
Instead of returning the errors directly, the new method `#getValidationResult()` (previously `#getValidationErrors()`, see #269) will return a
result object which can be used to retrieve the errors in different ways.

- `#all()`: returns all errors
- `#mapped()`: returns a mapped version of the errors
- `#useFirstErrorOnly()`: determines that `#all()`/`#mapped()` will work with only the first error occurrence for each param
- `#throw()`: throws an `Error` with the same API as the result object

Tasks:
- [x] Develop the features
- [x] Update docs
- [x] Find a more suitable name instead of `#all()` (we're not really dealing with all errors sometimes...)

Closes #235, closes #223, closes #30 

---

**Bonus:** I reorganized lots of stuff in the docs!

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -5,6 +5,7 @@
     "sub": true,
     "curly": true,
     "eqeqeq": true,
+    "expr": true,
     "indent": 4,
     "latedef": "nofunc",
     "newcap": true,
--- a/README.md
+++ b/README.md
@@ -48,11 +48,11 @@ app.post('/:urlparam', function(req, res) {
   // OR find the relevent param in all areas
   req.sanitize('postparam').toBoolean();
 
-  // Alternatively use `var errors = yield req.getValidationErrors();`
+  // Alternatively use `var result = yield req.getValidationErrors();`
   // when using generators e.g. with co-express
-  req.getValidationErrors().then(function(errors) {
-    if (errors) {
-      res.send('There have been validation errors: ' + util.inspect(errors), 400);
+  req.getValidationErrors().then(function(result) {
+    if (!result.isEmpty()) {
+      res.send('There have been validation errors: ' + util.inspect(errors.array()), 400);
       return;
     }
     res.json({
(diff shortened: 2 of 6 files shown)


IOAyman:
Very good job, @gustavohenke 👍  Thank you!

PR #289, opened 2016-12-01
Add TypeScript type definition

IOAyman:
# Summery
Mainly, I have added the TypeScript type definitions for this library (based on the definitions at [DefinitelyTyped/express-validator](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e2af6d0/express-validator/express-validator.d.ts), which are out-dated and incomplete).

I am submitting this PR here because:
- The definitions would be included with the library itself, no need to install an extra package, for instance `@types/express-validator`.
- The process of merging fixes over at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) is slow and my take a while.

_PS: I will also submit the same update to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)._

### Other minor changes:
- Upgraded validator.js to [v6.2.0](https://github.com/chriso/validator.js/blob/6.2.0/CHANGELOG.md).
- Added missing `unescape` sanitizer.
- Updated some JSDoc.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,6 +28,10 @@ declare module "express-validator" {
 declare namespace ExpressValidator {
 
   interface ValidatorFunction { (item: string | {}, message?: string): Validator; }
+  /**
+   * This one's used for RegexRoutes
+   * @see https://github.com/ctavan/express-validator#regex-routes
+   */
   interface ValidatorExtraFunction extends ValidatorFunction { (matchIndex: number, message?: string): Validator; }
   interface SanitizerFunction { (item: string): Sanitizer; }
   interface Dictionary<T> { [key: string]: T; }
@@ -67,7 +71,7 @@ declare namespace ExpressValidator {
     validate: ValidatorExtraFunction;
     check: ValidatorExtraFunction;
     checkBody: ValidatorFunction;
-    checkFiles: ValidatorFunction;
+    checkCookies: ValidatorFunction;
     checkHeaders: ValidatorFunction;
     checkParams: ValidatorFunction;
     checkQuery: ValidatorFunction;
@@ -78,6 +82,7 @@ declare namespace ExpressValidator {
     sanitizeQuery: SanitizerFunction;
     sanitizeParams: SanitizerFunction;
     sanitizeHeaders: SanitizerFunction;
+    sanitizeCookies: SanitizerFunction;
 
     /**
      * @deprecated
@@ -277,7 +282,7 @@ declare namespace ExpressValidator.Options {
   }
 
   interface IsIntOptions extends MinMaxExtendedOptions {
-    allow_leading_zeroes: boolean;
+    allow_leading_zeroes?: boolean;
   }
 
   /**
@@ -311,10 +316,10 @@ declare namespace ExpressValidator.Options {
    * }
    */
   interface IsURLOptions {
-    protocols?: string[];
+    protocols?: string[]; // TODO Protocol type
     require_tld?: boolean;
     require_protocol?: boolean;
-    require_host: boolean;
+    require_host?: boolean;
     require_valid_protocol?: boolean;
     allow_underscores?: boolean;
     host_whitelist?: (string | RegExp)[];
@@ -345,8 +350,8 @@ declare namespace ExpressValidator.Options {
    * }
    */
   interface IsISSNOptions {
-    case_sensitive: boolean
-    require_hyphen: boolean
+    case_sensitive?: boolean
+    require_hyphen?: boolean
   }
 
   /**
@@ -405,16 +410,16 @@ declare namespace ExpressValidator.Options {
    * }
    */
   interface NormalizeEmailOptions {
-    all_lowercase: boolean
-    gmail_lowercase: boolean
-    gmail_remove_dots: boolean
-    gmail_remove_subaddress: boolean
-    gmail_convert_googlemaildotcom: boolean
-    outlookdotcom_lowercase: boolean
-    outlookdotcom_remove_subaddress: boolean
-    yahoo_lowercase: boolean
-    yahoo_remove_subaddress: boolean
-    icloud_lowercase: boolean
-    icloud_remove_subaddress: boolean
+    all_lowercase?: boolean
+    gmail_lowercase?: boolean
+    gmail_remove_dots?: boolean
+    gmail_remove_subaddress?: boolean
+    gmail_convert_googlemaildotcom?: boolean
+    outlookdotcom_lowercase?: boolean
+    outlookdotcom_remove_subaddress?: boolean
+    yahoo_lowercase?: boolean
+    yahoo_remove_subaddress?: boolean
+    icloud_lowercase?: boolean
+    icloud_remove_subaddress?: boolean
   }
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,7 +14,8 @@ declare namespace Express {
 declare module "express-validator" {
 
   /**
-   * @middlewareOptions see: https://github.com/ctavan/express-validator#middleware-options
+   * @param options see: https://github.com/ctavan/express-validator#middleware-options
+   * @constructor
    */
   function ExpressValidator(options?: ExpressValidator.ExpressValidatorOptions): express.RequestHandler;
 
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -17,8 +17,9 @@
     "url": "git://github.com/ctavan/express-validator.git"
   },
   "main": "./index.js",
+  "types": "./index.d.ts",
   "scripts": {
-    "test": "nyc mocha",
+    "test": "nyc mocha && tsc",
```
gustavohenke: Is there anything that is compiled by `tsc` here?

package.json
```diff
@@ -29,9 +30,12 @@
     "node": ">= 0.10"
   },
   "dependencies": {
+    "@types/bluebird": "~3.0.36",
+    "@types/express": "~4.0.34",
     "bluebird": "^3.4.0",
     "lodash": "^4.16.0",
-    "validator": "~6.1.0"
+    "typescript": "~2.0.10",
```
gustavohenke: This should definitely go to `devDependencies` otherwise everyone will get `typescript` installed

package.json
```diff
@@ -29,9 +30,12 @@
     "node": ">= 0.10"
   },
   "dependencies": {
+    "@types/bluebird": "~3.0.36",
+    "@types/express": "~4.0.34",
```
gustavohenke: Just to be clear: I think they should probably be moved to `devDependencies` (if this works in production as well)

IOAyman: Those must remain in `dependencies`. Otherwise the user would have to install them manually.

gustavohenke:
All good, but what about adding #290 as soon as it gets accepted?

IOAyman:
@gustavohenke Sure!

gustavohenke:
So it's your turn again :) I just merged #290.

IOAyman:
The structure of the schema object that is passed to `req.check()`, or any of its variants, is not defined; It just has to be an object.
So #290, does not affect this PR 😉 
I'll try to work on defining an interface for the schema object as well some other time :)

index.d.ts
```diff
@@ -0,0 +1,420 @@
...
+   */
+  function ExpressValidator(options?: ExpressValidator.Options.ExpressValidatorOptions): express.RequestHandler;
+
+  export = ExpressValidator;
+}
+
+// Internal Module.
+declare namespace ExpressValidator {
+
+  interface ValidatorFunction { (item: string | {}, message?: string): Validator; }
```
gustavohenke: This interface is used for specialized `checkSomething` methods; there should be an overload of it to include the schema validation possibility.

IOAyman: You mean an overload that accepts _'only the schema'_ object :) You're right. Will do.

IOAyman:
I have just added the type definition for validation schema, and taken into account the changes from #290 :)

gustavohenke:
Great @IOAyman, thanks!

Do you plan on refactoring something else? Any missing TODOs here or I'm good to merge?

index.d.ts
```diff
@@ -0,0 +1,517 @@
...
+  export type AlphanumericLocale = 'ar' | 'ar-AE' | 'ar-BH' | 'ar-DZ' | 'ar-EG' | 'ar-IQ' | 'ar-JO' | 'ar-KW' | 'ar-LB' | 'ar-LY' | 'ar-MA' | 'ar-QA' | 'ar-QM' | 'ar-SA' | 'ar-SD' | 'ar-SY' | 'ar-TN' | 'ar-YE' | 'cs-CZ' | 'da-DK' | 'de-DE' | 'en-AU' | 'en-GB' | 'en-HK' | 'en-IN' | 'en-NZ' | 'en-US' | 'en-ZA' | 'en-ZM' | 'es-ES' | 'fr-FR' | 'fr-BE' | 'hu-HU' | 'nl-BE' | 'nl-NL' | 'pl-PL' | 'pt-BR' | 'pt-PT' | 'ru-RU' | 'sr-RS' | 'sr-RS@latin' | 'tr-TR' | 'uk-UA'
+  export type MobilePhoneLocal = 'ar-DZ' | 'ar-SA' | 'ar-SY' | 'cs-CZ' | 'de-DE' | 'da-DK' | 'el-GR' | 'en-AU' | 'en-GB' | 'en-HK' | 'en-IN' | 'en-NZ' | 'en-US' | 'en-CA' | 'en-ZA' | 'en-ZM' | 'es-ES' | 'fi-FI' | 'fr-FR' | 'hu-HU' | 'it-IT' | 'ja-JP' | 'ms-MY' | 'nb-NO' | 'nn-NO' | 'pl-PL' | 'pt-PT' | 'ru-RU' | 'sr-RS' | 'tr-TR' | 'vi-VN' | 'zh-CN' | 'zh-TW'
+  export type Location = 'body' | 'params' | 'query' | 'headers' // TODO add cookies if #292 is accepted
```
IOAyman: I'll either add or remove this depending on the conclusion of #292.

test/typeDefinitionTest.ts
```diff
@@ -0,0 +1,142 @@
...
+
+
+  req.checkQuery('param').isEmail({ require_tld: true });
+  req.checkQuery('param', 'message');
+  req.checkQuery('param.child', 'message');
+  req.checkQuery(['param', 'child'], 'message').optional();
+  req.checkQuery(schema);
+
+
+  // TODO chain all validators
```
IOAyman: I'll add this in few minutes

index.d.ts
IOAyman: However, this TODO can be left as is for the moment.

gustavohenke:
Also some jobs on the Travis build failed. I just pushed a small change in the `.travis.yml` to update npm before anything else.
It's probably good to merge it, so the builds will always pass.

IOAyman:
Okey, I see! Fair point. I'll remove that contributors section, although it is quite nice to have it (maybe I'll push a script to automate the process of maintaining it some other time :wink:).

The latest build failure on this PR is probably due to [this TravisCI incident](https://www.traviscistatus.com/incidents/s45ml12bpzhl).

IOAyman:
One more thing, [I assumed](https://github.com/ctavan/express-validator/pull/289/files#diff-04c6e90faac2675aa89e2176d2eec7d8R522) this will be shipped in v3.1.0.
But in fact, I think these changes might be breaking somethings (that's what I mentioned in the README).
So, [according to SemVer](https://github.com/mojombo/semver/blob/master/semver.md#summary), I think it should be a major version bump.
What do you think @gustavohenke?

gustavohenke:
> I think these changes might be breaking somethings

Because of the supposed `.d.ts` conflicts, right?
If that's the case, I don't know.. we are not introducing breaking changes to express-validator itself.

I found this at [the TypeScript docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html):

> If you use an `import "foo"` statement, for instance, TypeScript may still look through `node_modules` & `node_modules/@types` folders to find the foo package

If this is part of the specification, then we're good to ship this PR.

IOAyman:
This does not break library itself, yes. But I think it would break the TS compilation, therefore the whole project build.

Yes, I've read that part.
But I think if a module's type definitions are shipped both in the module itself and are installed in `node_modules/@types`, TypeScript would use the ones in `@types`, if I am not mistaken.

IOAyman:
**Quick test:**  I've just tested that on in a project I already have set-up;
Installed both `@types/express-validator@2.20.33` and `express-validator@3.0.0` (from this branch).
`tsc` throws this:
`.../requestvalidation.ts(21,9): error TS2339: Property 'getValidationResult' does not exist on type 'Request'.`

The error disappears if I remove the `@types` dependency.

gustavohenke:
So it doesn't matter what, if you have a conflicting installation of `@types/express-validator`, it will not work 😄 

Therefore I think we should proceed with v3.1.0.

IOAyman:
Yes, the compilation wont pass 😄 I'll push the modifications in a few minutes with the 3.1.0 :) Let's get this merged.

gustavohenke:
Thanks.
Before merging I'll rebase and squash some of your commits to make this PR terse in the Git log.

gustavohenke:
Thanks again for everything @IOAyman!

IOAyman:
It is a pleasure @gustavohenke, giving back to the community 😄 
I'll be updating it whenever the library is updated :)

gustavohenke:
Landed in v3.1.0! 🚢

IOAyman:
Thanks! 👍

package.json
Spoki4: I checked many repositories and everywhere @types dependencies places in devDependencies. I think if I use only es5 or es6 why I need install types for typescript?

Issue #188, opened 2015-11-26
Custom validators - change execution context of function to "validator" object


fogine:
I would really appreciate if the `this` keyword of custom validator function would be bind to `validator's` object context on which it's executed.
By `validator` object I mean an object returned by `req.assert()` for example.

rustybailey:
Not near a computer and won't be able to look at this today...

However, if you remember, what is "this" currently bound to? Just so I 100% understand, can you give an example of what logging "this" would return after your proposed change?

fogine:
````javascript
{ 
  errorFormatter: [Function],
  param: 'name',
  value: 'david',
  validationErrors: [],
  failMsg: 'Invalid name',
  req:[Object],
  lastError: null
}
````

**Comment by rustybailey** [comment]
@ctavan thoughts?

**Comment by ctavan** [comment]
@fogine I have to admit that I still don't fully get your point. Can you extend your example a bit?

**Comment by fogine** [comment]
Well, methods `checkBody()`, `checkParams()`, `checkQuery()` and `assert()` on `req` object return all new `validator` instance, is that right?
So I would like to gain access to that object through `this` keyword in custom validator's function:

````javascript
app.use(expressValidator({
 customValidators: {
    isArray: function(value) {
        console.log(this);
        return Array.isArray(value);
    }
 }
}));
````

The console log from the above example would output:

````javascript
{ 
  errorFormatter: [Function],
  param: 'name',
  value: 'david',
  validationErrors: [],
  failMsg: 'Invalid name',
  req:[Object],
  lastError: null
}
````

-> if called like eg: `req.assert('name', 'Invalid name').isArray();`

Like so you get access to `req` object and `validationErrors`, `_validationErrors` lists from within an validator function.

**Comment by gustavohenke** [comment]
Accessing the `req` object is possible with the check API introduced in v4:
https://github.com/ctavan/express-validator#customvalidator

With this you can also check the errors of previous chains by using `validationResult(req)`:

```js
app.get('/foo', [
  check('bar')
    .isEmail()
    .custom((value, { req }) => {
      // If 'bar' is not an email, you will never know here.
      // isEmail() is set in the same validation
      validationResult(req).throw();
    }),
  check('baz').custom((value, { req }) => {
    // If 'bar' is not an email, this method will throw,
    // because the previous chain is done
    validationResult(req).throw();
  })
])
```

PR #298, opened 2016-12-24
Added custom function object calling with details.

auser:
Added the ability to use the `req` object in custom validators with README and test updates.

Addresses #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,44 @@ Use them with their validator name:
 req.checkBody('users', 'Users must be an array').isArray();
 req.checkQuery('time', 'Time must be an integer great than or equal to 5').isInt().gte(5)
 ```
+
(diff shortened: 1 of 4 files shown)


gustavohenke:
I'm quite afraid of this PR... it feels like is a breaking change.
Any custom validators accepting a rest argument will likely break/won't be possible because of this.

gustavohenke:
[Version 4.0.0 arrived](https://github.com/ctavan/express-validator/releases/tag/v4.0.0) and custom validators will now have access to the `req` object 😉 

Check [the README for complete docs](https://github.com/ctavan/express-validator/tree/v4.0.0) and the [upgrade guide for version 3 users](https://github.com/ctavan/express-validator/blob/v4.0.0/UPGRADE_GUIDE.md).

PR #299, opened 2016-12-26
Added check for missing validation method

maxsummers:
Added check for missing validation method, when validating schema object. This will throw user-friendly error instead of cryptic one: `TypeError: Cannot read property 'apply' of undefined`.


lib/express_validator.js
```diff
@@ -441,6 +441,10 @@ function validateSchema(schema, req, loc, options) {
       if (opts != null && !Array.isArray(opts)) {
         opts = [opts];
       }
+
```
IOAyman: There's an extra trailing spaces here causing the tests to fail. Please remove them

PR #306, opened 2017-01-14
Add default export in type definitions to support ES6 import syntax.

btraut:
By changing "export = ExpressValidator" to "export default ExpressValidator", the module can be imported using ES6 syntax:

Instead of forcing users to use `import expressValidator = require('express-validator');`, they can now use `import expressValidator from 'express-validator';`.

I ran into this issue initially when trying to move to all ES6 syntax. It also led me to https://github.com/DefinitelyTyped/DefinitelyTyped/issues/10742 which seems to suggest that ES6 syntax isn't supported.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,7 +21,7 @@ declare module "express-validator" {
    */
   function ExpressValidator(options?: ExpressValidator.Options.ExpressValidatorOptions): express.RequestHandler;
 
-  export = ExpressValidator;
+  export default ExpressValidator;
 }
 
 // Internal Module.


gustavohenke:
@IOAyman, thoughts?

kacepe:
It doesn't fix it for me.

index.d.ts
```diff
@@ -21,7 +21,7 @@ declare module "express-validator" {
    */
   function ExpressValidator(options?: ExpressValidator.Options.ExpressValidatorOptions): express.RequestHandler;
 
-  export = ExpressValidator;
+  export default ExpressValidator;
```
IOAyman: I don't think that's possible.

This lib [exports](https://github.com/ctavan/express-validator/blob/841944e40b5bbc48aa5a92287ccb8436b26e9e5c/lib/express_validator.js#L567) a default function that initializes it, but ES6 modules cannot directly export callable functions. This is the reason why TS forces you to import modules the CommonJS way (`import x = require('module')` ).

gustavohenke: And why are the tests passing then?

IOAyman: The way the lib is [imported](https://github.com/ctavan/express-validator/blob/841944e40b5bbc48aa5a92287ccb8436b26e9e5c/test/typeDefinitionTest.ts#L1) in the test is "generic": It works for all sorts of modules.

But if you try to do it the ES6 way `import ev from 'express-validator'` it would fails, even if you `export default ExpressValidator`

gustavohenke:
I think this PR is now superseded since #352 landed.

PR #308, opened 2017-02-03
#nullable()

Vladislao:
Use the nullable() method to allow null values.

```javascript
req.checkBody('price').nullable().isDecimal();
// if there is no error, req.body.price is either null or a valid decimal.
```

or even

```javascript
req.checkBody('price').optional().nullable().isDecimal();
// if there is no error, req.body.price is either undefined or null or a valid decimal.
```

Before it was like this
```javascript
if (req.body.price !== null) {
  req.checkBody('price').optional().isDecimal();
}
```

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -159,6 +159,14 @@ var expressValidator = function(options) {
     return this;
   };
 
+  ValidatorChain.prototype.nullable = function() {
+    if (this.value === null) {
+      this.skipValidating = true;
+    }
+
+    return this;
+  };
+
   ValidatorChain.prototype.withMessage = function(message) {
     if (this.lastError) {
       if (this.lastError.isAsync) {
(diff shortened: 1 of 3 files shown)


gustavohenke:
`.optional({ checkFalsy: true })` should suffice for this.

Let me know if you have other use cases.

Vladislao:
Hello, @gustavohenke
Falsy would allow me to pass any falsy value, like `false` or `''`. As for my example this is not a desired behaviour. Why would i store false as a price? :)
I can't say for everyone, but i think it's quite common to expect some value to be nullable, not just falsy

gustavohenke:
It's the contrary:

> If you want to skip validation based on the property being falsy (null, undefined, etc), you can pass in `{ checkFalsy: true }`

gustavohenke:
Okay, will reconsider, and perhaps think over this idea a little bit more.

deksden:
Imho, this is valuable addition.

When you make some api with optional parameter - you can expect that this parameter can be omitted, or if param it present - it should be checked strictly.

Maybe we should add some generic logic operator like:
* oneOf([array of checks]) - would be ok if at least one of checks is ok, 
* anyOf([array of checks]) - would be ok if any check is ok,
* (?) noneOf([array of checks]) - would be ok, if none of checks are ok (maybe not().any() will be same)

gustavohenke:
Up-to-date PR created: #441

PR #309, opened 2017-02-08
fix(type-definition): Import error, errorFormatter

kacepe:
- Fixed import error: `Module '"express-validator"' resolves to a non-module entity`
- Changed errorFormatter returned value.
- Add RequestValidation Interface on to Global Express's Request Interface.


gustavohenke:
cc @IOAyman

PR #315, opened 2017-03-01
Add optionalNil function

jensengar:
There are 2 reasons for doing this rather than using the existing `optional`. First, doing `optional({checkFalsy: true})` is too long/tedious. Secondly, `!this.value` does not handle the case when `value === false`. By that I mean I only want to skip validation if `value` is null or undefined but not if the value is `false` or `''`.

BTW, thanks for doing this repo, this is great.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "Gustavo Henke <guhenke@gmail.com>",
     "Ayman Nedjmeddine <theycallmethedr@gmail.com>"
   ],
-  "version": "3.1.2",
+  "version": "3.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "Gustavo Henke <guhenke@gmail.com>",
     "Ayman Nedjmeddine <theycallmethedr@gmail.com>"
   ],
-  "version": "3.1.2",
+  "version": "3.2.0",
```
gustavohenke: Please undo this. Version bumps are managed by the project collaborators

PR #320, opened 2017-03-12
Move @types from devDependency to dependency section

Spoki4:
I install only dependecies to production server, but I installed @types. I think it's bad.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,13 @@
     "node": ">= 0.10"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


Spoki4:
I checked many repositories and everywhere @types dependencies places in devDependencies. I think if I use only es5 or es6 why I should install types for typescript?

gustavohenke:
Do you mind finding some examples?
I'm fine changing this behavior for v4.

PR #337, opened 2017-04-19
Reduce module state to improve testability

ChaseMoskal:
Hello!

At the company I work for called INgrooves, we use `express-validator` in several applications — we thank all contributors of this project for an excellent tool!

Recently, we uncovered a bug in express-validator, which this PR fixes.
 - each of our apps used express-validator and some `customValidators`
 - we merged a few of these apps together into a single app, and experienced breakage — because `express-validator` keeps its `customValidators` in module-level state — the custom validators for route B overwrite the custom validators for route A, etc

This PR fixes the problem by simply moving this global state within the express validator constructor's scope
 - thus now each call to express validator has its own private ValidationChain prototype to work with (also internalized Sanitizer and the validateSchema function)
 - subsequent express validator calls won't contaminate previous calls

I say this improves testability, but in our case, the fix was necessary for our application to run (not tests) — this fix will certainly help anybody using `express-validator` in an async-testing situation.

### DIFF
--- a/test/testabilityTest.js
+++ b/test/testabilityTest.js
@@ -1,0 +1,61 @@
+
+var chai = require('chai');
+var express = require('express');
+var request = require('supertest');
+var expressValidator = require('../index');
+
+describe('Testability:', function() {
+  describe('validator works with multiple express apps,', function() {
+
+    // This test case:
+    //  - creates two express apps
+    //  - initializes each app with different custom validators of the same name
+    //  - verifies that custom validators declared for one app do not
+    //    contaminate the other app's validators
+    it('custom validators do not contaminate other sessions', function(done) {
+      var happyApp = express();
+      var angryApp = express();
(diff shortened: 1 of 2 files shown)


test/testabilityTest.js
```diff
@@ -0,0 +1,61 @@
...
+
+    // This test case:
+    //  - creates two express apps
+    //  - initializes each app with different custom validators of the same name
+    //  - verifies that custom validators declared for one app do not
+    //    contaminate the other app's validators
+    it('custom validators do not contaminate other sessions', function(done) {
+      var happyApp = express();
+      var angryApp = express();
+      chai.assert(happyApp !== angryApp, 'express app is not a singleton');
```
gustavohenke: I would say that this is irrelevant, as you're testing express itself

gustavohenke:
Released in [v3.2.0](https://github.com/ctavan/express-validator/releases/tag/v3.2.0) 🚢

PR #342, opened 2017-05-13
Updated dependecies

hendrikschneider:


gustavohenke:
Since bluebird is already using `ˆ` operator, updating it is not needed (the typing dependency was updated as well).
For validator dep, I'm not willing to update it for now, since it introduces a breaking change that could require us to do a major release as well.

PR #348, opened 2017-05-31
declare trim parameter as optional in type definitions

nicolashenry:
trim, ltrim, and rtrim should be usable without parameter using typescript.


gustavohenke:
Thanks!

Issue #350, opened 2017-06-10
Unable to use express-validator with TypeScript 2.3.4


taicho:
Based on how the typings are written (the ones that ship with express-validator). I am unable to import the library without compile-time errors.

I've tried:
import * as expressValidator from 'express-validator';
import {ExpressValidator} from 'express-validator';

Both of which result in:
Module '"express-validator"' resolves to a non-module entity and cannot be imported using this construct.'

Is anybody else running into this issue? I figure a workaround would be to delete index.d.ts from node_modules/express-validator/ but I feel concerned about this issue coming up again when the next release of express-validator is published...

Thanks in advance!

baptistelambert:
I had the same problem and tried importing like so:
```ts
import expressValidator = require('express-validator')
```

Not the prettiest syntax but it works on my side.

taicho:
@baptistelambert Thanks. :) Guess I'll have to go the same route too. ;-/

therealTP:
Importing each individual method from 'express-validator/check' works for me:
```
import { check, body, query, param, validationResult } from 'express-validator/check';
``` 
You then have access to each method within the file.

PR #351, opened 2017-06-11
REF #350 Fix for TypeScript Typings

taicho:
Without this fix users of TypeScript (at least 2.3.4) unable to import
the library using standard “import” syntax without either a) removing
the typings or b) using “require”.

Fix for #350


taicho:
#350

PR #352, opened 2017-06-12
REF #350 Fix for TypeScript Typings Revisted

taicho:
This has been a bad day. This is tested both in project tests and in
real project and should resolve the issues with imports in TypeScript.

Fixes: #350

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,16 +10,17 @@
 declare namespace Express {
   interface Request extends ExpressValidator.RequestValidation {}
 }
+
+// External express-validator module.
+declare module "express-validator" {
   import express = require('express');
 
   /**
    * @param options see: https://github.com/ctavan/express-validator#middleware-options
    * @constructor
    */
-  declare function ExpressValidator(options?: ExpressValidator.Options.ExpressValidatorOptions): express.RequestHandler;
-
-  export = ExpressValidator;
-
+  export function ExpressValidator(options?: ExpressValidator.Options.ExpressValidatorOptions): express.RequestHandler;  
+}
 // Internal Module.
 declare namespace ExpressValidator {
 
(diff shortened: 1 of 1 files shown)


gustavohenke:
cc @IOAyman - do you mind reviewing?

taicho:
@gustavohenke My pleasure! I was driven by my desire to use this great library. :) The problem for TypeScript users like myself was that if using the recommended "import" syntax to try and reference/require the library it would result in compile-time errors preventing compilation altogether. The only alternatives were to either delete the definitions entirely so TS was essentially blind to the library or use "require"; both of these options resulted in TS being unaware of the typings rendering them useless. As you can see from my multiple attempts yesterday I am not the most competent typings author not to mention I wasn't testing properly and was a little impatient. That being said this should completely resolve this issue for TS users who want to use your library. On a side note, I'm uncertain if this broke due to recent updates to TS, what I am certain about is that without this fix the library was unusable in the current version (2.3.4) of TS.

emauricio:
Hi there, quick question about in this version how to declare the custom validators?
in the prev. version i was able to do like 
```
declare global {
    namespace ExpressValidator {      
        interface Validator {
           //here the custom stuff
...
```
but now is not working, any suggestions, already try to google it, but no much about this topic, thx in advance.

versions
 - typescript: 2.4.1 
 - express-validator: 3.2.1

nicolashenry:
I use this in a .d.ts file: 
```typescript
import 'express-validator';
declare module 'express-validator' {
	interface Validator {
		//here the custom stuff
	}
}
```

gustavohenke:
Thanks for the help, @nicolashenry!
I just added a test for this: a8dde3138b5edabfe12c318404f14c1d78205e32

emauricio:
Ohh thank you very much 😄

PR #354, opened 2017-06-13
Fix `isValid.then` check bug in case validator returns undefined/null

cotko:
I wanted to use the `isUUID` validator. I've set the version as option, like so:
```js
isUUID: {
  options: {
    version: 4
  }
}
```
it turned out, that validator expects the version as a second param instead:

https://github.com/chriso/validator.js/blob/master/lib/isUUID.js#L26

so the version was `{version: 4}` and the line
```js
  var version = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 'all';

```
didn't fall back to `all` correctly. Now when it tried to find the pattern, it got undefined:

```js
  var pattern = uuid[version]; // uuid[{version: 4}] does not exist
  return pattern && pattern.test(str);
```

this lib then threw error as it did `undefined.then` [here](https://github.com/ctavan/express-validator/blob/master/lib/express_validator.js#L494).

This PR fixes such issues, however I don't know if it might be better to leave it throw an error so that developer finds out what's actually wrong.

### DIFF
--- a/lib/express_validator.js
+++ b/lib/express_validator.js
@@ -18,6 +18,23 @@ var warnAsyncValidationErrors = deprecate(
   'req.asyncValidationErrors() may be removed in a future version. Use req.getValidationResult() instead.'
 );
 
+/**
+ * display warnings once per each validator
+ * which returns null or undefined as a validation
+ * result
+ */
+var warnValidatorNilReturn = (function() {
+  var warned = {};
+  return function(methodName, returnedValue) {
+    if (warned[methodName]) {
+      return;
+    }
+    warned[methodName] = true;
+    console.warn('WARNING: unexpected return value: `' + returnedValue + '` returned by `' + methodName + '` validator');
+  }
+
+}());
+
 // When validator upgraded to v5, they removed automatic string coercion
 // The next few methods (up to validator.init()) restores that functionality
 // so that express-validator can continue to function normally
@@ -493,8 +510,8 @@ function makeValidator(methodName, container) {
 
     var isNilValue = isValid === undefined || isValid === null
 
-    if(isNilValue) {
-      warnValidatorNilReturn(methodName, isValid, this.value);
+    if (isNilValue) {
+      warnValidatorNilReturn(methodName, isValid);
     }
 
     if (!isNilValue && isValid.then) {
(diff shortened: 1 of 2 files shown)


gustavohenke:
The `options` of a validator must be provided either as an array or as a single value.
Since we add no layer on top of validator.js validators, we don't know what the name of the options are.

Therefore, what you should have done is

```js
isUUID: {
  options: 4
  // or
  options: [ 4 ]
}
```

gustavohenke:
Since a falsy return from the validator means such field is invalid, then I can't merge your changes.

gustavohenke:
Actually, thinking better about your change, it makes sense!

cotko:
I might have written a bit bad description of the problem in a PR :)

I've opened it because this lib first handled validation so that it only checked truthy/falsy returns from validators, but it later gained support for promises:

https://github.com/ctavan/express-validator/commit/6e13c901bf5e0dea5b0085b1366c9a6c6dae9956#diff-6c47d849cc78d146191d4d361699b452R226

By not checking if `isValid` is *null* or *undefined* before checking for `isValid.then` it means that node validator's validatros can make this lib throw errors and possibly exit the process, which is the case of the *isUUID*; [if this line](https://github.com/chriso/validator.js/blob/master/lib/isUUID.js#L26) would read

```js
return !!(pattern && pattern.test(str));

// instead of 

return pattern && pattern.test(str);

```
and if all other validators would also guarantee to not return undefined/null as falsy tests, the then devs could get this error only by using their *customValidators*, but since express validator is bundled with those validators, I think this should be handled.

But just silently ignoring such cases would make it harder for devs to find out the misconfigured validator, which would be my case with the uuid check. I was thinking about this and if you're ok with it, I can update the PR so that if *isValid* is null or undefined, it would log a warning telling which validator returned unexpected value and add a flag to turn this warnings off, similar to how other libs like [bluebird](http://bluebirdjs.com/docs/api/promise.config.html)/[eventemitter](https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n) etc. show non critical messages.

gustavohenke:
> I can update the PR so that if isValid is null or undefined, it would log a warning telling which validator returned unexpected value

This seems like a good idea!

> add a flag to turn this warnings off

I recently implemented a warning in deprecated methods; this deprecation message is only shown once (otherwise each request would say the same thing, over and over again):
https://github.com/ctavan/express-validator/commit/addbecd58e1e9a5240da3271f5e24f8f72a8c546

Perhaps this is something along the lines of what you imagined?

cotko:
Yes, something like this, I'll use it and update the PR :)

gustavohenke:
ping @cotko - are you willing to fix this soon?

cotko:
I've updated the PR. Warnings are printed once per validator name which returns invalid value (didn't use util.deprecate as it prints 'deprecated' which is a bit misleading:) )

PR #366, opened 2017-07-09
Move to unit tests

gustavohenke:
This is a try I'm giving to make the tests easier to read, understand and write.
From another point of view, this could even make the codebase easier to refactor.

Current approach makes quite difficult to know what the intended behaviour of each tests is, and the setup is a bit cumbersome, requiring us to do [stuff like cleaning up `require.cache`](https://github.com/ctavan/express-validator/blob/v3.2.0/test/checkTest.js#L56-L57).
Basically, it's a kind of integration tests.

The new approach is to unit-test instead.
We create mocks of the express request and check the express-validator API after we decorate it/validate it.


gustavohenke:
👆 The coverage is even better than before (let's just hope the functionality keeps the same after a few rounds of upcoming refactors...)

PR #418, opened 2017-09-16
Sanitizers in the validation chain API

### LINKED ISSUE #417: How to use sanitize with V4 API?
I have spent way more time than I will admit attempting to overcome a very basic edit of a form field, how to make sure that an entry is not empty or just contains blanks using the recommended V4 style.  This issue has been raised previous issues, including #402 and #397.  It appears that using V4 the only options are: 
1. Including other libraries (which are already used in express-validator) to do this.
2. Use a combination of V3 and V4 validation.
3. Create a custom validation method for every input field.
4. Add sanitation to the check method. 
5. Just use validator without express-validator.

Which ever method is used should be clearly documented with a working example.

gustavohenke:
Adds support for using validator sanitizers in the validation chain API.

Example:

```js
check('some.field').trim().escape().not().isEmpty();
```

Closes #417

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ app.post('/user', [
     // You can customize per validator messages with .withMessage()
     .isEmail().withMessage('must be an email')
 
+    // Every sanitizer method in the validator lib is available as well!
+    .trim()
+    .normalizeEmail()
+
     // ...or throw your own errors using validators created with .custom()
     .custom(value => {
       return findUserByEmail(value).then(user => {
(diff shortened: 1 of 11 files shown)


wannymiarelli:
Using (on v 4.2.1)
`check('price').toFloat()`
the price value in the body is not parsed to float val. None of the sanitize methods works. No error reported.
If i switch to sanitize API it works well.

gustavohenke:
Hi @wannymiarelli, see relevant discussion in #438.

PR #422, opened 2017-09-19
Fix onlyFirstError attribute name on UPDATE-GUIDE

tuliobraga:
The right attribute name is 'onlyFirstError' instead of 'firstErrorOnly'


gustavohenke:
Nice catch!

PR #440, opened 2017-09-29
Custom validator is missing in the typescript definition file

markus-perl:
It was not possible to use the custom validator in a Typescript project, because the declaration for this validator was missing in the shared-typings.d.ts file.


gustavohenke:
Thanks for the heads up!

gustavohenke:
Well, actually, that was my mistake.

I imagine you're using the legacy API.
In this case, we can't simply add TS support for `custom` there because the same implementation is not there.

The check API, however, had support for TS from the start.

markus-perl:
Thank you Gustavo for pointing out, that I was using the legacy API. I upgraded to the new API and everything works as expected.

PR #441, opened 2017-09-30
Optional - nullable option

gustavohenke:
Adds support for `nullable` option in the `.optional()` method of the validation chain.
When `true`, this option will skip validation from fields with `null` or `undefined` values.

Original idea by @Vladislao #308


deksden:
Great news.

Should we expect this PR to be merged into 4.2.2/4.3.0? Any timeframe?

gustavohenke:
As this is a feature, you may expect it for 4.3.0.
I haven't merged it yet because I'm sorting out if I want to fix any bugs before releasing a new minor version. But that would be soon, eg 1 week or less.

PR #453, opened 2017-10-13
Implemented separated matched data for different locaitons

### LINKED ISSUE #447: can matchedData return just the body validated params?
Hi there.
I'm having a little trouble with mathedData, let me explain.
```javascript
const validateConfig = [
  body(['id', 'description'])
    .exists().withMessage('must be specified')
    .not().isEmpty().withMessage('cannot be empty')
    .custom(value => {
      if (typeof value === 'string') return true
      throw new Error(`must be a 'string'`)
    })
]

const validateTypeParam = [
  param('type')
    .isIn(['foo', 'bar']).withMessage(`must be either 'foo' or 'bar'`)
]

router.post('/:type', validateConfig, validateTypeParam,
  (req, res, next) => {
    const errors = validationResult(req)
    if (!errors.isEmpty()) {
      return res.status(422).json({ errors: errors.mapped() })
    }
    let test = matchedData(req)
    /** At this point I'd like to have test equal to
       {
           id: 'foo', // validated req.body.id
           description: 'bar' // validated req.body.description
       }

       And instead I'm getting
       {
           id: 'foo', // validated req.body.id
           description: 'bar' // validated req.body.description,
           type: 'foo' //validated req.params.type
       }
     **/
    db.then((db) => {
      return db.get(`tests.${req.params.type}`).push(test).write()
    }).then((data) => {
      res.json(data)
    }).catch((err) => next(err))
  })
```

So, is there any way that I can return only the 'req.body' validated params?
Currently i'm just doing ``` delete test.type ``` but I'd like to have it cleaner.

Thanks in advance

villamide:
Implemented `locations` in matchedData to be able to get matched data from specified fields (like body, params...)

Closes #447

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,15 +180,30 @@ These methods are all available via `require('express-validator/filter')`.
 
 ### `matchedData(req[, options])`
 - `req`: the express request object.
-- `options` *(optional)*: an object of options. Defaults to `{ onlyValidData: true }`
+- `options` *(optional)*: an object which accepts the following options:
+  - `onlyValidData`: if set to `false`, the returned value includes data from fields
+    that didn't pass their validations. Defaults to `true`.
+  - `locations`: an array of locations to extract the data from. The acceptable values include
+    `body`, `cookies`, `headers`, `param` and `query`. Defaults to `undefined`, which means all locations.
 > *Returns:* an object of data validated by the `check` APIs.
 
 Extracts data validated by the `check` APIs from the request and builds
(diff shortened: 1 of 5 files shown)


gustavohenke:
Very nice first contribution on GitHub, @villamide! 😄 

As you will find out, I changed a few things. But your implementation is awesome, thanks!

gustavohenke:
Published in v4.3.0 🚢

Issue #427, opened 2017-09-20
Global errorFormatter not used


jplusje:
When I've provided an errorFormatter function during initialization, it's not always being used?

working example (the output is indeed being mapped by the provided errorFormatter function):
```
req.getValidationResult().then(result => {
  console.log(result.mapped());
});
```

not working example (the output is being formatted by the default errorFormatter function):
```
import * as expressValidatorCheck from "express-validator/check";

try {
  expressValidatorCheck.validationResult(req).throw();
} catch (err) {
  return res
    .status(422)
    .json({ errors: err.mapped() })
    .end();
}
```

Is this a bug or am I using this incorrectly?

gustavohenke:
The check API doesn't have this concept of a default error formatter yet.
For now, you must call `.formatWith()` at every validation result.

Suggestions on how to implement this are welcome!

insightcoder:
Would love to see this implemented as well.

alrav:
Hello there @gustavohenke - I've created a proposal for default `validation-result` options - starting with error formatter: https://github.com/ctavan/express-validator/pull/474

Let me know if it's of interest to you 👍

PR #474, opened 2017-11-20
check: support default options for validationResult

alrav:
This is a proposal to resolve https://github.com/ctavan/express-validator/issues/427, by introducing a way to set default options for the `validation-result` module

It introduces a new `.withDefaults` function, which will return a new instance of the module with specified default options. The decision to return a new instance was to avoid affecting the global module state in the case that multiple `validationResult` implementations are desired within the same application.

This is loosely inspired on the concept of `request.defaults` from the [request/request](https://github.com/request/request#requestdefaultsoptions) module

Here's a brief example:
```
// baseValidationResult.js ----
const { validationResult } = require('express-validator/check');

const result = validationResult.withDefaults({
    formatter: (error) => {
        return {
            myLocation: error.location,
        };
    }
});

module.exports = result;

// within an express router ----
const validationResult = require('./baseValidationResult');

errors = validationResult(req);

if (!errors.isEmpty()) {
  return res.status(422).json({
    errors: errors.mapped(),
  });
}
```

Note: This is my first time getting acquainted with TypeScript, so let me know if the definitions make sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ try {
 
 ### `.withDefaults(options)`
 - `options` *(optional)*: an object of options. Defaults to `{ formatter: error => error }`
-> *Returns:* a new validation result instance with provided default options
+> *Returns:* a new [validationResult](#validationresultreq) function is returned, using the provided options
 
 This is useful when you have a consistent set of options you would like to use for all validation results throughout your application.
 
--- a/README.md
+++ b/README.md
@@ -367,6 +367,28 @@ try {
 }
 ```
 
+### `.withDefaults(options)`
+- `options` *(optional)*: an object of options. Defaults to `{ formatter: error => error }`
+> *Returns:* a new validation result instance with provided default options
+
+This is useful when you have a consistent set of options you would like to use for all validation results throughout your application.
+
+Below is an example which sets a default error formatter:
+
+```
+const { validationResult } = require('express-validator/check');
+
+const result = validationResult.withDefaults({
+    formatter: (error) => {
+        return {
+            myLocation: error.location,
+        };
+    }
+});
+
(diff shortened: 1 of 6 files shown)


check/type-definition.spec.ts
```diff
@@ -23,6 +24,15 @@ result.formatWith(error => {
   error.value;
 });
 
+withDefaults({
```
gustavohenke: Can you use `validationResult.withDefaults()`? So that it matches the documented example.
I think this will also require changes to how the type declaration is done...

gustavohenke: It will also be worth changing the test to `const result: ValidationResult = validationResult.withDefaults(...)(req)`. 
Should catch all possible typing bugs...

alrav:
Thanks for the feedback and absolutely! I'm working on addressing your requested changes 👍

alrav:
Hello @gustavohenke - comments addressed. Let me know if anything else needs a tweak.

README.md
```diff
@@ -367,6 +367,28 @@ try {
 }
 ```
 
+### `.withDefaults(options)`
+- `options` *(optional)*: an object of options. Defaults to `{ formatter: error => error }`
+> *Returns:* a new validation result instance with provided default options
```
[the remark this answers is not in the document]
alrav: Makes perfect sense to me

check/type-definition.spec.ts
alrav: I think it should be set now.

I kept `validationResult` interface as lower-camel-cased, since it might be more easily identifiable as a function. If any objections, I can also change to `ValidationResult`.

gustavohenke:
Thanks again!

Issue #444, opened 2017-10-04
.custom() not working with promise


jplusje:
I am trying to get .custom() working from the check API but it appears that it's not working?

Working (only showing essential information):
```
check.body('email').trim().isEmail().normalizeEmail().custom((_value) => {
    return true;
});

expressValidatorCheck.validationResult(req)..throw();
```

Not working (only showing essential information:
```
check.body('email').trim().isEmail().normalizeEmail().custom((_value) => {
    return Promise.resolve();
});

check.validationResult(req).throw();
```
The docs state:
_The custom validator may return a promise to indicate an async validation task. In case it's rejected, the field is considered invalid._
Therefore I would expect both examples to result in a "passed validation" however the second example always fails.

gustavohenke:
Thanks for the report @jplusje.

The code currently expects all validators to unbox as a truthy value. This means even async validators that return promises should have a value, eg `Promise.resolve(true)` to be considered as passed.

jplusje:
After changing my code to always resolve the promise but include a true/false it indeed works. I've also tested it with my actual implementation which returns a real promise and resolves it later and that also works.

thanks!

gustavohenke:
Yes, but as the docs say, *a resolved promise should mean a passed validation*, and this is not happening yet. Will fix it so this doesn't happen to other users in the future.

dimpen:
This happened to me too. It needs to `resolve(true)`. You have to mention it in the API documentation until you change it otherwise it's very confusing. BTW I don't think it's a bug IF you mention it in the docs.

gustavohenke:
@dimpen that's not the intended behaviour.

PR #478, opened 2017-12-07
README.md: add a precision for the usage of .custom

antonin-lebrard:
The README.md actually says  

>  and return falsy values to indicate the field is invalid.

But this is the only mention of it, at the end of a sentence, and is not actually that clear when referring to the first example in the Usage part

I had to launch my test in debug mode to see the custom validator checking if a result is provided at the end of a custom validation


gustavohenke:
This behaviour was fixed. As commented in #444, it's not the intended thing to happen.

PR #481, opened 2017-12-17
Add custom sanitizer

tranvansang:
Add custom sanitizer

Example
```js
router.get('/:number',
  check('number').toInt()
  .customSanitizer((value, mul, add, {path, location, req}) => value * mul + add), mul, add),
  nextHandler)
```

I also implementing async sanitizer to make something like

```js
router.get('/user/:user',
  check('user')
  .customSanitizer(user => new Promise((resolve, reject) => User.findById(user, (err, user) =>
    err ? reject(err) : resolve(user)
  ))), handler)
```
in [my another branch](https://github.com/tranvansang/express-validator/tree/async-sanitizer) (still in progress)


filter/sanitize.js
```diff
@@ -24,11 +24,21 @@ module.exports = (fields, locations) => {
...
         sanitizers.push({
           sanitizer: sanitizerFn,
+          custom: false,
           options
         });
         return middleware;
       };
     });
 
+  middleware.customSanitizer = (sanitizer, ...options) => {
```
gustavohenke: To keep the same interface as the `.custom()` validator method, this should not receive any options.

check/select-fields.js
```diff
@@ -60,7 +60,9 @@ function expand(object, path, paths) {
 
 function createSanitizerMapper({ sanitizers = [] }) {
   return field => sanitizers.reduce((prev, sanitizer) => {
-    const value = typeof prev.value === 'string' ?
+    const value = sanitizer.custom ? (prev.value === undefined ?
+      undefined : sanitizer.sanitizer(prev.value, ...sanitizer.options, prev)
+    ) : typeof prev.value === 'string' ?
```
gustavohenke: Can you please indent this block better?

check/check.js
```diff
@@ -77,6 +78,15 @@ module.exports = (fields, locations, message) => {
     return middleware;
   };
 
+  middleware.customSanitizer = (sanitizer, ...options) => {
```
gustavohenke: To keep the same interface as the `.custom()` validator method, this should not receive any options.

README.md
```diff
@@ -276,6 +276,22 @@ app.post('/create-user', [
 ], loginHandler);
 ```
 
+### `.customSanitizer(sanitizer, [options])`
+- `sanitizer(value, ...options, { value, location, path })`: the custom sanitizer function.
```
gustavohenke: Should take `req` to follow the `.custom()` validator interface

README.md
```diff
@@ -276,6 +276,22 @@ app.post('/create-user', [
 ], loginHandler);
 ```
 
+### `.customSanitizer(sanitizer, [options])`
```
gustavohenke: Can you also add it to the `filter` API (as it exists there as well)?

PR #484, opened 2017-12-25
docs : updated when and how to use wildcards(*).

### LINKED ISSUE #414: wildcards docs
Where can i look at documentation for wildcards?
Is there some other shortcuts `*`, i need something that can check if array contains at least one item with property and value i need

VaibhavAjayGupta:
docs : updated when and how to use wildcards(*). Also added the note that validation and sanitization are different.

I think many people are combining the code of request sanitization with validation. Which is not a persistent way of sanitizing the request. And also are unaware that wildcards(*) exist.

Closes #414


README.md
```diff
@@ -391,6 +405,13 @@ module.exports = result;
 
 ---
 
+## Validation and Sanitization are different
+
+Validation and Sanitization of a request are two different process. Do them seprately. Although all the sanitizer methods canbe used during validation but those sanitizations will be effective for the validation purpose only. That sanitization will not persist. 
+For persistent sanitization of request it's must to use 'express-validator/filter' methods.
```
gustavohenke: I saw many cases of people getting astonished with such behaviour. It definitely needs to be changed (as a breaking change), so I'm not keen on adding it to the docs now that the next release is due soon.

README.md
```diff
@@ -233,6 +233,20 @@ Same as `sanitize(fields)`, but only sanitizing `req.query`.
 
 ---
 
+## Wildcards (*)
+To validate or sanitize an array of strings use wildcards(*).
```
gustavohenke: This sentence is not necessarily the only case to use wildcards, so mentioning something like "nested fields" would be better.

README.md
```diff
@@ -233,6 +233,20 @@ Same as `sanitize(fields)`, but only sanitizing `req.query`.
...
+
+```js
+// Wildcards * are accepted!
```
gustavohenke: Add comment saying that the `postalCode` of every item under `addresses` will be validated?

README.md
VaibhavAjayGupta: Cool.

README.md
VaibhavAjayGupta: oops.

README.md
VaibhavAjayGupta: Cool.

VaibhavAjayGupta:
@gustavohenke Thank you for your feedback. I have made the necessary changes.

Issue #458, opened 2017-10-18
Wildcard support for subfields seems ignoring undefined fields


xuefengwang:
Here is my test case for this scenario. The code is supposed to trigger missing addresses[1].address1, but it actually passed validation. I'm using express-validator 4.2.1.

Route handler:
```
router.post('/', [
    check('addresses.*.address1').exists(),
    check('addresses.*.city').exists()
], (req, res, next) => {
    const errors = validationResult(req);
    if (!errors.isEmpty()) {
        console.log(errors.mapped());
        res.status(400).json({errors: errors.mapped()});
    } else {
        res.status(200).json({title: 'Express'});
    }
});
```

Test case:
```
describe('index', function () {
    it('should throw invalid address', (done) => {
        chai.request('http://localhost:3000')
            .post('/')
            .send({
                'addresses': [ { 'address1': '123 St', 'city': 'Test' }, { 'city': 'Test' } ]
            })
            .end((err, res) => {
                res.should.have.status(400);
                res.body.should.be.a('object');
                res.body.should.have.property('errors');
                res.body.errors.should.have.property('addresses[1].address1');
                res.body.errors['addresses[1].address1'].msg.should.equal('Invalid value');
                done();
            });
    });
});
```

Thanks,
William

xuefengwang:
After some debugging into the code, it seems in select-fields.js, as below
```
    if (instances.length > 1) {
      const withValue = instances.filter(field => field.value !== undefined);
      instances = withValue.length ? withValue : [instances[0]];
    }
```
addresses[1].address1 was in the instances. But after the logic above, since the value is undefined for addresses[1].address1, then it got filtered out.

gustavohenke:
Thanks for finding this out!
This if statement is there for cases when you have the same field in more than one location. In such cases, they work fine, but this also affects legitimate cases like yours.

It seems like a complex bug 😕

xuefengwang:
Thanks for looking into this, @gustavohenke. As you mentioned, the code is trying to get undefined from other locations. For the timing being, as a hack/workaround, I changed code locally as below to make it work. 
```
    if (instances.length > 1 && field.indexOf('*') === -1) {
      const withValue = instances.filter(field => field.value !== undefined);
      instances = withValue.length ? withValue : [instances[0]];
    }
```

gustavohenke:
Partial fix merged in #492.
[...]

PR #492, opened 2018-01-12
fix issue where undefined nested properties under a wildcard are not validated

torra:
this partially fixes the #458. as long as you specify where in the request to look for the field, instead of using the catch-call `check()` (ie, `param()`, `query()`, etc.), this fixes the issue.

it's only a partial fix, but it is a simple change and this at least makes it possible to do this kind of validation (ie checking for `undefined`'s)

### DIFF
--- a/check/select-fields.js
+++ b/check/select-fields.js
@@ -13,7 +13,7 @@ module.exports = (req, context) => {
       .filter(optionalityFilter)
       .value();
 
-    if (instances.length > 1) {
+    if (instances.length > 1 && context.locations.length > 1) {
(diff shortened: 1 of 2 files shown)


torra:
@xuefengwang @gustavohenke - fyi

@gustavohenke i would love some feedback if this is a bad idea for other reasons

@xuefengwang this would solve the problem in your example (in #458) if you use `body()` instead of `check()`

torra:
thanks for merging @gustavohenke - mind cutting a new release for npm?

PR #495, opened 2018-01-16
Proof of concept: break validation chain early.

neobyteUK:
This is a brief proof of concept PR to demonstrate returning early from a validation chain when a single validator fails. Sometimes you only want to call into a complex (expensive) business logic validator if simpler validators have already passed.

Best explained by an example:
```
body('submitted_id', null, {break_chain_on_failure: true})
  .trim()
  .isUuid()
  .custom(value => someExpensiveDBCall(value))
  .custom(value => someExpensiveNetworkCall(value))
```
For example, this will fail with a single UUID validator error if the submitted value is not a UUID. It will not attempt any of the further logic tests. If the UUID check succeeds but the DB check fails, the network check will not be performed.

Any thoughts on the idea? It doesn't introduce obvious breaking changes but I haven't put a lot of thought into edge cases.

I'm also relatively new to express-validator so maybe there is a way to accomplish the same thing nicely that I have overlooked?

### DIFF
--- a/check/check-body.js
+++ b/check/check-body.js
@@ -1,2 +1,2 @@
 const check = require('./check');
-module.exports = (fields, message) => check(fields, ['body'], message);
+module.exports = (fields, message, opts) => check(fields, ['body'], message, opts);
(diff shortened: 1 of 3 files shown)


neobyteUK:
Ok great I'll flesh out the idea with the .bail() interface

PR #516, opened 2018-02-05
Custom Sanitizer

### LINKED ISSUE #445: Custom sanitizer support?
Will there be a way to add custom sanitization function? Something likes this:
`sanitize('foo').custom(customSanitizer)`

gustavohenke:
Closes #445, closes #481

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,23 @@ app.get('/', sanitizeBody('trimMe').trim(), (req, res, next) => {
 
 Any of the sanitization methods listed by [validator.js](https://github.com/chriso/validator.js) are made available in all sanitization chains created by express-validator, as long as we're supporting the most up-to-date validator version.
 
+Additionally, the following methods are also available:
+
+### `.customSanitizer(sanitizer)`
+- `sanitizer(value, { req, location, path })`: the custom sanitizer function.  
+  Receives the value of the field being sanitized, as well as the express request, the location and the field path.
+> *Returns:* the current sanitization chain instance
+
+Adds a custom sanitizer to the current sanitization chain. It must synchronously return the new value.
+
+Example:
+
+```js
+app.get('/object/:id', sanitizeParam('id').customSanitizer((value, { req }) => {
+  return req.query.type === 'user' ? ObjectId(value) : Number(value);
+}), objectHandler)
+```
+
 ---
 
 ## Validation Chain API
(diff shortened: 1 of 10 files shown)


pkyeck:
@gustavohenke tried to use a `customSanitizer()` for the first time today and was wondering it is called twice? Once before and once after validation.

```javascript
'use strict';

const { check } = require('express-validator/check');

module.exports = check('filter')
  .optional()
  .matches(/^[0-9,]+$/i)
  .withMessage('filter must be a list of IDs')
  .custom(arr => {
    console.log('custom validator', arr);
    return arr.length <= 20;
  })
  .withMessage('filter can only contain a max of 20 IDs')
  .customSanitizer(value => {
    console.log('custom sanitizer', value);
    const dedupe = Array.from(new Set(value.split(',')));
    return dedupe.map(item => parseInt(item, 10)).filter(item => !Number.isNaN(item));
  });
```

outputs

```bash
custom sanitizer 2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,
custom validator [ 2, 3, 5 ]
custom sanitizer 2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,2,2,3,5,
```

pkyeck:
was still running 5.0.0. updated but I'm still seeing the two sanitizer logs but both with the original input and not like in one of the linked issues, a double sanitization

pkyeck:
just FYI: after the update other unit-tests involving express-validator are failing :( 

```bash
Expected value to equal:
  {
    "eventId": {
      "location": "body",
      "msg": "id has to be int bigger 0",
      "param": "id",
      "value": NaN
    }
  }
Received:
  {
    "eventId": {
      "location": "body",
      "msg": "id has to be int bigger 0",
      "param": "id",
      "value": "                      "
    }
  }
```

the value on error has changed. so kind a should have been a major update, right!?
... but too late for that now.

gustavohenke:
> updated but I'm still seeing the two sanitizer logs but both with the original input and not like in one of the linked issues, a double sanitization

I will have to ask for you to create an issue then, with all steps to reproduce it.

As for the latter, yeah, that is a two-sided coin. It's a breaking change when one is checking responses like you are, but it's also a bug fix as in the real submitted value is the one in the error objects.
Sorry for it, I think 😞

pkyeck:
@gustavohenke thanks. will check about creating a new issue. just a quick question: when should the sanitizer usually be called? before or after validation?

gustavohenke:
_before_ validation, always

PR #521, opened 2018-02-12
Fix! Type Sanitizer

Klemele:
Howdy guys,

Seems to be an issue with the last release and the typing from 'Sanitizer'.
This quick 'revert' seems to fix it ;).


gustavohenke:
Fixed it myself :)

PR #536, opened 2018-03-06
Fix missing values from matchedData when using arrays in oneOf

### LINKED ISSUE #533: MatchedData not returning all values after validating passed with oneOf
This problem started after upgrading from 4.3.0 to 5.0.x.
When sending json data for validation, matchedData does not return last value:

Example:
```
{ "body":"human", "version":"test" }
```

Validation:
```
router.post("/test", [
	oneOf([
		body("body").exists(),
		body("body").isEmpty()
	]),
	oneOf([
		body("version").exists(),
		body("version").isEmpty()
	])
], (req, res) => {
	const bodyValues = matchedData(req);

	console.log(bodyValues);
```

Would return only:
```
{ "body":"human" }
```
It works perfectly fine with 4.3.0 returning:
```
{ "body":"human", "version":"test" }
```

Tried also the way as it was documented and matchedData returned empty object.

Hopefully oneOf is ment to work in array as it used to, since it covers 4 different options for body values in the first example. This second way I would need to write all 4 separately.

```
router.post("/testi",
	oneOf([
		[
			body("version").exists(),
			body("body").exists()
		],
		[
			body("body").isEmpty(),
			body("version").isEmpty()
		]
	]), (req, res) => {
		const bodyValues = matchedData(req);

		console.log(bodyValues);
```

jonathonlui:
Fixes #533.

If `oneOf` has a successful chain then any error that match the first successful chain's fields is filtered out of `req._validationErrorsOneOf` so that `matchedData` won't exclude it.

### DIFF
--- a/check/one-of.js
+++ b/check/one-of.js
@@ -8,18 +8,24 @@ module.exports = (validationChains, message) => (req, res, next) => {
     return Array.isArray(chain) ? chain.map(getContext) : getContext(chain);
   });
 
+  // These promises resolve to an array with results with the chain that
+  // generated the results
   const promises = validationChains.map(chain => {
     if (!Array.isArray(chain)) {
-      return run(chain);
+      return run(chain).then(results => [results, chain]);
     }
 
-    return Promise.all(chain.map(run)).then(results => _.flatten(results, true));
+    return Promise.all(chain.map(run)).then(results => [_.flatten(results, true), chain]);
   });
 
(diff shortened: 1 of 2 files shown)


Zakini:
Any progress on this? Anything I can help with?

gustavohenke:
🚢 v5.1.1 is out with the fix!

PR #539, opened 2018-03-13
docs: method name typo at README.md

muratsu:
Small typo fix. The correct method name is `isUUID` according to `validator` package.


gustavohenke:
Thanks!!

PR #546, opened 2018-03-23
Added SanitizersSchema

sant123:
This PR adds a Sanitizer schema when using the function `checkSchema`. 

Sanitizers like `toInt`, `escape` and so on now appears in the intellisense section and Typescript yields no more.

If something is missing please let me now.

### DIFF
--- a/check/schema.d.ts
+++ b/check/schema.d.ts
@@ -1,5 +1,6 @@
 import { ValidationChain, ValidatorOptions, CustomValidator } from './check';
 import { Location } from './location';
+import { CustomSanitizer } from '../filter/sanitize';
 
 export function checkSchema(schema: ValidationSchema): ValidationChain[];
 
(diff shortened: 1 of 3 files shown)


check/schema.d.ts
```diff
@@ -72,4 +78,20 @@ interface ValidationParamSchema {
   isVariableWidth?: ValidatorSchemaOptions;
   isWhitelisted?: ValidatorSchemaOptions;
   matches?: ValidatorSchemaOptions;
-}
\ No newline at end of file
+}
+
+interface SanitizersSchema {
+  blacklist?: SanitizerSchemaOptions;
```
gustavohenke: might need to add `customSanitizer`

check/schema.d.ts
```diff
@@ -10,16 +10,22 @@ type ValidatorSchemaOptions<T = any> = true | {
...
 
   custom?: ValidatorSchemaOptions<CustomValidator>;
   exists?: ValidatorSchemaOptions;
   optional?: boolean | ValidatorOptions.OptionalOptions;
+}
 
-  equals?: ValidatorSchemaOptions;
+interface ValidatorsSchema {
   contains?: ValidatorSchemaOptions;
+  equals?: ValidatorSchemaOptions;
```
gustavohenke: 🙇 thanks!

check/schema.d.ts
sant123: Added.

PR #548, opened 2018-03-23
Sanitization fixes

sant123:
+ Now the sanitizers run after validations
+ The error summary is no longer touched by sanitizers (for better explanation when a validation fails)

For this schema

``` js
{
    id: {
        in: ["params"],
        isNumeric: { errorMessage: "The 'id' param must be a number" },
        toInt: true,
    },
}
```
When `id` is not numeric, the sanitizer produced this:

``` js
{
    "id": {
        "location": "params",
        "param": "id",
        "value": NaN,
        "msg": "Invalid value"
    }
}
```

And turning this into JSON produced:

``` js
{
    "id": {
        "location": "params",
        "param": "id",
        "value": null,
        "msg": "Invalid value"
    }
}
```

With this PR the error summary `.mapped()` remains untouched, however in the `request` object these values remain been sanitized (that is the goal of the library).

+ When using schema, sanitizers no longer overrides errorMessage from validations.

Using the same schema above, `msg` said `Invalid value` while there is a `errorMessage` set:

``` js
isNumeric: { errorMessage: "The 'id' param must be a number" }
```

This PR preserves the errorMessage.

Additional notes:

If only sanitizers are provided, if `id` is not an int then `req.params.id` will be `NaN`. 

``` js
{
    id: {
        in: ["params"],
        toInt: true,
    },
}
```

### DIFF
--- a/utils/select-fields.js
+++ b/utils/select-fields.js
@@ -1,10 +1,13 @@
 const _ = require('lodash');
 const formatParamOutput = require('./format-param-output');
 
-module.exports = (req, context, sanitizeFields = true) => {
+module.exports = (req, context, config) => {
+  const defaults = { sanitizeFields: true };
+  config = Object.assign({}, defaults, config);
+
   let allFields = [];
   const optionalityFilter = createOptionalityFilter(context);
-  const sanitizerMapper = sanitizeFields ? createSanitizerMapper(req, context) : e => e;
+  const sanitizerMapper = config.sanitizeFields ? createSanitizerMapper(req, context) : e => e;
 
   context.fields.forEach(field => {
     let instances = _(context.locations)
--- a/utils/select-fields.js
+++ b/utils/select-fields.js
@@ -1,10 +1,10 @@
 const _ = require('lodash');
 const formatParamOutput = require('./format-param-output');
 
-module.exports = (req, context) => {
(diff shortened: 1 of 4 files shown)


utils/select-fields.js
```diff
@@ -1,10 +1,10 @@
 const _ = require('lodash');
 const formatParamOutput = require('./format-param-output');
 
-module.exports = (req, context) => {
+module.exports = (req, context, sanitizeFields = true) => {
```
gustavohenke: Can you please make this parameter an object? So to avoid [boolean traps](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).

sant123: Added :)

gustavohenke:
Ugh, there is a breaking change here. The [Validation Chain docs](https://github.com/ctavan/express-validator#validation-chain-api) says:

> If you use any of the sanitizers together with validators, the validated value is the sanitized one.

Sorry I didn't notice it before!

I have, however, fixed the error message setting, but in a quite different way though. Please check the closing commit to see it.

sant123:
Wonderful, you solved it with a better approach. May you upload these changes to NPM? Thank you.

gustavohenke:
🚢 v5.1.0 released with this

PR #564, opened 2018-04-10
fix "exists" can not customize message when using schema validation

WeiAnAn:
fix "exists" can not custom message when using schema validation.


gustavohenke:
Published [`v5.2.0`](https://github.com/express-validator/express-validator/releases/tag/v5.2.0) 🚢

PR #587, opened 2018-06-03
Update custom validator callback syntax

errietta:
It seems that (at least as of version 4.3.0) it's not enough to just reject the promise or throw an error if there's an issue with the validation. The validator needs to resolve or return a truthy value, otherwise an error will be shown to the user.


gustavohenke:
This was fixed with [v5.0.0](https://github.com/express-validator/express-validator/releases/tag/v5.0.0)/#444. It should be safe to update unless you are relying on some non-standard behaviour.

If the docs for v4.x ever get ported back to this new Docusaurus format, let's make sure to include this change 👍

PR #603, opened 2018-07-04
Small correction to the sanitizer docs example

knight-of-zero:
`id` is the name of the URL param being sanitized... but the variable in the function is named `value`.


gustavohenke:
Thanks!

Issue #552, opened 2018-03-29
Support multiple custom validators/sanitisers with schemas


gustavohenke:
We can't use multiple custom validators/sanitisers when we use schemas, as an object can't have multiple keys with the same name.

This makes it impossible to break them into smaller pieces of validation/sanitisation.
Example taken from #550: one that does only email DNS validation, another one that does only email availability validation, etc.

The only way I can imagine for now is to accept an array of custom validators.
Any other suggestions are welcome.

Parsec01:
I would like to be able to have such a possibility as well

OmgImAlexis:
For those that need a work around for now this should help.

```js
/**
 * Runs an array of validators over a value
 */
const validationArray = validators => {
	let errors = [];

	return {
		custom: {
			errorMessage: () => {
				return errors.length >= 1 ? errors.join('\n') : 'value is invalid';
			},
			options: (value, opts) => {
				errors = [];
				const results = validators.map((validator, index) => {
					logger.debug(`validating ${value} [${index + 1}/${validators.length}]`);
					const result = validator.validate(value, opts);
					// If validation failed set current Error
					if (result === false) {
						const errorMessage = validator.errorMessage || `${value} is invalid`;
						logger.debug(`setting error to ${errorMessage}`);
						errors.push(errorMessage);
					}
					return result;
				});
				return results.every(value => value === true);
			}
		}
	};
};
```

To use it pass an array of functions that return a validate field and optionally an `errorMessage` field.

```js
/**
 * Ensures field value is from provided array
 */
const oneOf = array => ({
	errorMessage: `value must be one of [${array.join(', ').replace(/, ([^,]*)$/, ' or $1')}]`,
	validate: value => array.includes(value)
});

/**
 * Ensures field value is boolean
 */
const isBoolean = {
	validate: value => typeof value === 'boolean'
};

/**
 * Schema
 */
export const schema = {
	type: {
		in: ['params'],
		...validationArray([
            isBoolean,
			oneOf([
				'virtual',
				'ipv4',
				'ipv6'
			])
		])
	}
};
```

sobiodarlington:
You can do multiple validations in the "custom" prop

```js

function isValidPhone() {
	
}

function anotherValidation() {
	
}

Validation.testSchema= {
	'user.phone': {
        in: ['body'],
        custom: {
            options(value, { req, location, path }) {
                if (!isValidPhone(value)) {
                    this.message = 'Invalid phone';
                    return false;
[...]

Issue #527, opened 2018-02-16
checkSchema - custom validator and exists errorMessage


andrey-hohlov:
Hi!

I'm trying to use Schema Validation with 5.0.0.

1. `custom` validator throw validation error `validatorCfg.validator is not a function` 
2. For `exists` param `errorMessage` ignored

```javascript
checkSchema({
  title: {
    in: ['body'],
    exists: {
      errorMessage: 'title is required', // <- doesn't work
    },
    custom(value) { // <- doesn't work
      console.log(value);
      return true;
    }
  },
  password: {
    isLength: {
      errorMessage: 'Password should be at least 7 chars long', // <- works
    },
  },
});
```

`customValidators` param and all in this doc section https://github.com/ctavan/express-validator#legacy-api are deprecated...

What am I doing wrong?

gustavohenke:
Hm, the `exists` one is strange, by "doesn't work" do you mean the error message is still `Invalid value`? It shouldn't be, so there might be something else wrong with your setup.

The `custom` one should be specified by passing `options`.
I will add an example to help clear future questions about it.

This is a very minimal app with both validators working:
```js
const { checkSchema, validationResult } = require('./check');

const express = require('express');
const app = express();
app.use(express.json());
app.post('/*', checkSchema({
  foo: {
    in: ['body'],
    exists: {
      errorMessage: 'bla'
    },
    custom: {
      options: value => (value || '').startsWith('foo')
    }
  }
}), (req, res) => {
  res.json(validationResult(req).array());
});

app.listen(3000);
```

andrey-hohlov:
@gustavohenke thank you!

It's all right with `exists`, my mistake.
Also, I tried `custom` with async function and it works too.

Is there a way to use multiple custom validators with different error messages?
Like:
```javascript
check('password')
    .custom(value => someValidationFunction(value))
    .withMessage('Some validation error message')
    .custom(value => anotherValidationFunction(value))
    .withMessage('Another validation error message')
```

gustavohenke:
If you don't use schema, then yes.

andrey-hohlov:
@gustavohenke same story for custom sanitizers?

Schema better for using, chains ugly :(

gustavohenke:
Yes, the same for custom sanitizers.

mohsincynexis:
not working with async.js here is my code:
`email: {
    errorMessage: 'Please enter a valid email address',
    isEmail : true,
    trim: true,
    custom: {
[...]

PR #604, opened 2018-07-04
Add createCheckAPI

prettymuchbryce:
As discussed in #527, adds a new public method `createCheckAPI` which allows a user to initialize a new instance of `check`, `checkSchema`, etc using `customValidators` and `customSanitizers`. This greatly improves the usability of the check API. It also happens to solve #552. The goal of this change is to be completely backwards compatible.

TODO
- Additional unit tests
- Typescript definitions
- Documentation

@gustavohenke Would love to get your input on the direction so far.


check/create-check-api.js
```diff
@@ -0,0 +1,33 @@
...
+  customValidators,
+  customSanitizers
+} = {}) => {
+  const customCheck = (fields, locations, message) =>
+    check(
+      fields,
+      locations,
+      message,
+      Object.assign({}, customValidators, validator),
+      Object.assign({}, customSanitizers, validator)
```
gustavohenke: By doing this, one could have their custom validators/sanitizers overwritten by validator's ones.
Could be an astonishment point for users that deliberately wanted to overwrite the standard ones, or for those that didn't know one with the same name exists and so their function never runs.

check/validation-chain-builders.js
```diff
@@ -1,6 +1,6 @@
-const check = require('./check');
+const checkModule = require('./check');
 
-const buildCheckFunction = locations => (fields, message) => check(fields, locations, message);
+const buildCheckFunction = locations => (fields, message, check = checkModule) => check(fields, locations, message);
```
gustavohenke: Did you mean to pass `check` to the outer function? `(locations, check = checkModule) => ...`?

check/check.js
```diff
@@ -17,10 +17,10 @@ module.exports = (fields, locations, message) => {
     }, next);
   };
 
-  Object.keys(validator)
+  Object.keys(validatorSet)
```
gustavohenke: When `createCheckAPI()` mixes standard validators with custom ones, the custom ones won't be built with the flag `custom: true`.
Internally, this is used by to determine how to call the validator (or sanitizer).

A better option is to always add the standard ones from validator.js, and change `validatorSet`  to simply `customVaildators`, then extend them like it's done a few lines below with `isArray`/`isString`/`exists`.

check/validation-chain-builders.js
prettymuchbryce: I'm not sure I understand the question. We need to use our custom version of check in the case that the user uses `createCheckAPI` so it is passed in as an argument. For the normal case, we just use the default argument checkModule for the normal non-custom check module.

If you look at `create-check-api.js` this is where we pass in our custom check function to `buildCheckFunction`.

prettymuchbryce:
@gustavohenke Any recommendations on how to gracefully test this? I have a couple of tests, but ideally we run all of the same tests that touch the exports of `index` with those from `createCheckAPI` instead. I could copy and paste all of the relevant tests, but it seems like a little too much code duplication.

fedeci (commented):
Hi @prettymuchbryce, this is definitely a smart solution, but should be reworked to fit the updated codebase😬
Can you handle this?

PR #622, opened 2018-08-11
note that custom sanitizers only work with strings

kevinsperrine:


codfish:
@kevinsperrine @gustavohenke is this the expected behavior though? Not sure why that would be, we should probably be looking to get https://github.com/express-validator/express-validator/pull/632 merged instead of this, right? FYI @tranvansang

kevinsperrine:
@codfish Absolutely the priority should be getting #632 should be merged; however, after spending many hours trying to figure out why it wasn't working and then finding this https://github.com/express-validator/express-validator/issues/601#issuecomment-404515156 that it has effectively been broken for 6 months I decided to open this, so that others wouldn't waste any time either.

tranvansang:
Hi, guys. Sorry for the criticism ideal. Don't you think express-validator's codebase is a bit difficult to be maintain while we have ES6 standard, everything can/should be promise.

I used to heavily using express-validator, spend hours if not days to get understand how does it work. I even wrote bunch of tests in my project just to verify how does express-validator works (e.g. empty string (`''`) DOES EXISTS (`chain.exists()`)). 

And eventually, I decided to wrote my own lib [express-transformer](https://www.npmjs.com/package/express-transformer), by around 150 lines of ES6 code for the basic functions. And other checkers are all built based on them.

gustavohenke:
Sanitizers now run with non-string types -- PR with the change: #711 🎉 

@tranvansang I would be very happy to have you contributing back into express-validator.
It is indeed an old library with lots of legacy use cases, so if you think we could improve it together, I would love it.

PR #628, opened 2018-08-18
Clarify which type to use for schema validation

Otto-AA:
Fixes https://github.com/express-validator/express-validator/issues/626


gustavohenke:
express-validator is written in TypeScript now, so this should not be a problem anymore.
Please let me know if you still face it.

PR #632, opened 2018-08-23
Update select-fields.js

### LINKED ISSUE #630: Custom sanitizer does not run on non-string type value
It looks like [this line](https://github.com/express-validator/express-validator/blob/master/utils/select-fields.js#L73) produces the bug

`body('page').customSanitizer(page => page - 1)`

with `req = {body: {page: 1}}` ends up with `req.body.page` retains being 1 after sanitized

tranvansang:
fix #630

### DIFF
--- a/utils/select-fields.js
+++ b/utils/select-fields.js
@@ -70,7 +70,7 @@ function expand(object, path, paths) {
 
 function createSanitizerMapper(req, { sanitizers = [] }, { sanitize = true }) {
   return !sanitize ? field => field : field => sanitizers.reduce((prev, sanitizer) => {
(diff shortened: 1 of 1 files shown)


alzeih:
I'd like to get this merged, so I've fixed the test it breaks over at #654 .

alzeih:
@gustavohenke I have now added tests for the existing (with strings) and new (other types) functionality at #654. I have also included this PR there to show the tests pass with this commit applied to the tree.

gustavohenke:
Hello! A fix should has been published in [`v6.0.0`](https://github.com/express-validator/express-validator/releases/tag/v6.0.0) 🚢, please try it out!

Issue #580, opened 2018-05-16
sanitizers does not run if the value is not a string


Bene-Graham:
I noticed the following when writing unit test for my middleware.
Take the following example:

Excerpt from body:
```
req.body = {
     isHappy: 1234
}
```
Excerpt from middleware array
```
checkBody("isHappy")
        .toBoolean(true)
        .not().isEmpty()
        .withMessage((value, { req, location, path }) => req.__("ValFieldIsRequired", req.__("isHappy")))
        .isBoolean()
        .withMessage((value, { req, location, path }) => req.__("ValFieldIsBoolean", req.__("isHappy"))),
```

The sanitized output is still 1234. When going through the documentation for validator it states under strict mode only 1 and true will return true all other values are false.

After a hour or so of debugging the express validator code I think I found the line responsible for what I am seeing 
https://github.com/express-validator/express-validator/blob/2cb4bcd7770329450b8e0880811d8390ebde41bd/utils/select-fields.js#L71

It looks like it by passes the santizer if the value passed in is not a string. I am wondering should this follow the same way the check function is done, by casting the value to a string?

gustavohenke:
Hello! A fix should has been published in [`v6.0.0`](https://github.com/express-validator/express-validator/releases/tag/v6.0.0) 🚢, please try it out!

PR #641, opened 2018-09-18
Allow Sanitization of Non String Params

Bene-Graham:
This should act like how the validator part of the library works, by toString() the value then perform the action against it

This should deal with my issue #580 

The one thing I was not sure on, is where to put the tests that the value values are passed and sanitized correctly. I ended up creating a sanitizer.spec.js in the check folder.


Bene-Graham:
I fixed the linting issues

gustavohenke:
Feel free to reply with your thoughts after my review above.
There's no need to start changing the code straight away 🙂

utils/select-fields.js
```diff
@@ -70,16 +71,14 @@ function expand(object, path, paths) {
 
 function createSanitizerMapper(req, { sanitizers = [] }, { sanitize = true }) {
   return !sanitize ? field => field : field => sanitizers.reduce((prev, sanitizer) => {
-    const value = typeof prev.value === 'string' ?
-      callSanitizer(sanitizer, prev) :
-      prev.value;
-
+    const value = !!prev.value ? callSanitizer(sanitizer, prev) : prev.value;
```
Bene-Graham: Not sure why I needed to a this check. If ended up just calling callSanitizer(sanitizer, prev) test would fail. Mainly because the prev.value is undefined.

Bene-Graham:
> I'm not sure if always converting the value back to string is the desired behaviour.
validator.js is unable to deal with non-string values, so that's fine for standard sanitizers, but maybe not for custom ones.

You are 100% right. Miss read the validator code in the runner.js file.

>Also, I'm not sure if the tests with the standard validators (toBoolean, toInt) are relevant; most of the behaviour is guaranteed by other means.

OK, I have removed the tests
---

I also added a comment to a line of code, that I needed to add to make the test pass.

alzeih:
I’m not sure converting to a string is the right approach for object and array like types?

If a new `customSanitizer` method (with a new name) was created that didn’t use the validator.js library then it could call the custom sanitizer code with whatever input, not just if the value is a string, and wouldn’t break existing code. It might mean that this new custom sanitizer couldn’t be chained with other sanitizers from validator.js though, or only can be chained effectively if it returns a string to other sanitizers.

Bene-Graham:
> I’m not sure converting to a string is the right approach for object and array like types?
> 
> If a new `customSanitizer` method (with a new name) was created that didn’t use the validator.js library then it could call the custom sanitizer code with whatever input, not just if the value is a string, and wouldn’t break existing code. It might mean that this new custom sanitizer couldn’t be chained with other sanitizers from validator.js though, or only can be chained effectively if it returns a string to other sanitizers. #632 addresses this.

If I am reading #632 correctly it won't solve the issue where the built in sanitizers won't run when the values are not already strings.

alzeih:
Yes, the second half of that comment is not relevant here (and is deleted).

String conversion for object and array like types though sounds like it could have some problems though to me?

gustavohenke:
Hello! A fix should has been published in [`v6.0.0`](https://github.com/express-validator/express-validator/releases/tag/v6.0.0) 🚢, please try it out!

PR #648, opened 2018-10-23
Add `defined` option for optional

Tol1:
Adds new option for `optional`: `defined`, which requires one to define given property. Should be used in combination of `checkFalsy` or `nullable`.
I made this, because REST principles declare that one should always send whole object on update and create, and explicity set optional values `null`. Previously this could be achieved only with oneOf: null or real validator.
Does not change default functionality.


Tol1:
> 
> 
> There's `exists`. Does that suit you?
> 
> https://express-validator.github.io/docs/validation-chain-api.html#existsoptions

It would, but I'd like that a default. Otherwise I need to include `exists` check on more than 99.9% of my validators, because most of my fields are required, and of optionals only small fraction can be undefined, mostly on query parameters.

fedeci (commented):
Hi, I think this can be easily achieved with the `exist()` validator and that adding this feature deviates from the intent of the validator.
/cc @gustavohenke

fedeci:
I'm closing this for now!

PR #654, opened 2018-11-03
Tests for customSanitizers with strings and other types

alzeih:
customSanitizers can (with another PR) run on other types of values including undefined, so the check must be more specific or the customSanitizer must check for undefined values.

This fixes the tests for #632 , but can be merged independently if acceptable and passes CI.

### DIFF
--- a/check/one-of.spec.js
+++ b/check/one-of.spec.js
@@ -1,5 +1,5 @@
 const { expect } = require('chai');
-const { check, oneOf } = require('./');
+const { check, cookie, oneOf } = require('./');
 
 describe('check: checkOneOf middleware', () => {
   it('returns errors from all chains', () => {
@@ -61,7 +61,7 @@ describe('check: checkOneOf middleware', () => {
 
     return oneOf([
       check('foo').trim(),
-      [check('foo').customSanitizer(value => value.toUpperCase())]
+      [cookie('foo').customSanitizer(value => value.toUpperCase())]
     ])(req, {}, () => {}).then(() => {
       expect(req.cookies.foo).to.equal('BAR');
     });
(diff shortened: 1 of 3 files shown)


alzeih:
Doe this make sense to the maintainers why I suggest this change? 

When `customSanitizer` functions can be run with any type it means customSanitizers can be run on `undefined` values, i.e: value is not sent with the request. As custom sanitizers may want to be run on `undefined` values (e.g: setting a form's checkbox value to boolean false if it isn't present), I think this is acceptable for `customSanitizer` functions.

What this means is custom sanitizers will be run on `undefined` values too now. Since this test only sets the cookie `foo`, when `check('foo')` is run, it tries to run the sanitizer in all places, e.g: on body, which has no `foo`. Calling `.toUpperCase()` on `undefined` breaks the tests.

I think silently failing to sanitize non strings is a worse existing behaviour than sanitizers loudly failing if they are run on `undefined` values, but I can see how both could be considered breaking changes in behaviour.

The alternate fix for this test with the other PR would be to change

```
[check('foo').customSanitizer(value => value.toUpperCase())]
```
to something like
```
[check('foo').customSanitizer( value => typeof value === 'string' ? value.toUpperCase() : value)]
```

Would this be a preferred fix?

alzeih:
Rebased new and fixed tests into one commit and changed PR title to reflect this.

alzeih:
Rebased for CI changes in PR #656

kolya182:
@gustavohenke Can you take a look at this, very helpful functionality to have

gustavohenke:
So I rewrote the whole codebase, and this is now supported -- thus I'm closing this PR 🙈
Implementation: #711

PR #667, opened 2018-12-07
.withData() support

recyclingben:
Allows a user to incorporate extra data with an error.

```javascript
        check("userId")
            .isInt().withMessage("userId could not be parsed to an integer").withData({ errorCode: 1 })
            .isLength(10).withMessage("userId was not length 10").withData({ errorCode: 2 })
```
This could even be used to send different http status codes depending on the errors in the future. The main purpose of this at the moment, though, would be to allow users of an API to programmatically discern between different errors as opposed to manually reading the error message or using a regular expression.

### DIFF
--- a/check/check.js
+++ b/check/check.js
@@ -90,11 +90,11 @@ module.exports = (fields, locations, message) => {
     return middleware;
   };
 
-  middleware.withData = (data, destructure = true) => {
+  middleware.withData = (data, spread = true) => {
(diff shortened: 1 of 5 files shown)


recyclingben:
Ah, that makes sense, @gustavohenke. Perhaps we could add this to documentation, then?

gustavohenke:
Sure thing! Created #758 to track this.

Sorry for the delay again!

PR #711, opened 2019-05-20
context-runners: sanitize non-string fields

### LINKED ISSUE #580: sanitizers does not run if the value is not a string

### LINKED ISSUE #630: Custom sanitizer does not run on non-string type value

### LINKED ISSUE #651: Bypass sanitize vulnerability
Bypass sanitize vulnerability:
	//app.js
	const express = require('express');
	const app = express();
	const { sanitizeQuery } = require('express-validator/filter');

	app.get('/', [sanitizeQuery('id').escape()],async(req,res)=>{
		res.send('id is ' + req.query.id);
	});

	app.listen(3000,()=>{
		console.log('3000');
	});

How To Bypass :

	before exploit:	
		if send request http://127.0.0.1:3000/?id=<h1>@meisamrce</h1>
		result like this : id is &lt;h1&gt;@meisamrce&lt;&#x2F;h1&gt;gt;
		you see express-validator escape all string

	Exploit:
		for bypass escape string just change the field to array 
		example : http://127.0.0.1:3000/?id[]=<h1>@meisamrce</h1>
		you see run your html or xss payload :)

gustavohenke:
**This is a breaking change**

Now non-string values can finally be sanitized, woo!
Closes #641, closes #632, fixes #630, fixes #580, fixes #651

### DIFF
--- a/src/context-runners/sanitize.spec.ts
+++ b/src/context-runners/sanitize.spec.ts
@@ -12,10 +12,10 @@ beforeEach(() => {
 it('maps instances using custom sanitizers in the context', async () => {
   const context = new Context([], []);
 
-  const sanitizer1 = jest.fn(value => value + '!!');
+  const sanitizer1 = jest.fn(value => value * 2);
   context.addSanitization(sanitizer1, { custom: true });
 
-  const sanitizer2 = jest.fn(value => 'wow, so much ' + value);
+  const sanitizer2 = jest.fn(value => value * 4);
   context.addSanitization(sanitizer2, { custom: true });
 
   const instances = await runner.run(req, context, [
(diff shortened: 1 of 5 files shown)


gustavohenke:
Also closes #654.

Issue #679, opened 2019-01-16
toArray() sanitizer


soulchild:
Given that there are sanitizers like `toBoolean()`, `toDate()`, etc. would it make sense to add a `toArray()` sanitizer as well? 

This would come in handy especially when dealing with multiple request parameters of the same name (for instance when using checkboxes) which are often transformed into an array, but will be left as a string if only one parameter was transmitted. (I guess this is a feature of express [body-parser](https://github.com/expressjs/body-parser) and similar middlewares…)

What do you guys think?

gustavohenke:
That's a great idea -- tagged it as an enhancement!

soulchild:
Cool! I could take a stab at it and submit a PR, if you like.

All current sanitizers seem to just delegate the hard work to the [validator](https://github.com/chriso/validator.js) library. Would it be okay to make an exception here and implement the `toArray()` functionality within [sanitizers-impl.ts](https://github.com/express-validator/express-validator/blob/master/src/chain/sanitizers-impl.ts) itself because `validator` does not offer it?

gustavohenke:
Yup, that would be it!

express-validator gained some "custom standard validators" on demand as well: https://github.com/express-validator/express-validator/blob/56153943ea10a0bc505f5f63c0367341ce1462be/src/chain/validators-impl.ts#L27-L33

Maybe you could follow suit?

PR #713, opened 2019-05-22
Add toArray() sanitizer (#679)

soulchild:
This pull request adds a toArray() sanitizer as discussed in #679.

### DIFF
--- a/src/chain/sanitizers-impl.spec.ts
+++ b/src/chain/sanitizers-impl.spec.ts
@@ -3,9 +3,10 @@ import * as validator from 'validator';
 import { SanitizersImpl } from './sanitizers-impl';
 import { Context } from '../context';
 import { Sanitizers } from './sanitizers';
+import { CustomSanitizer, Meta } from '../base';
 
 let chain: any;
-let context: Context;
+let context: jest.Mocked<Context>;
 let sanitizers: Sanitizers<any>;
 
 beforeEach(() => {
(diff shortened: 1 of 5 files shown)


src/chain/sanitizers-impl.spec.ts
```diff
@@ -103,3 +104,28 @@ describe('#customSanitizer()', () => {
...
+    sanitizers.toArray();
+
+    const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+    const toArray = context.addSanitization.mock.calls[0][0] as CustomSanitizer;
+
+    expect(toArray([], meta)).toEqual([]);
+    expect(toArray('foo', meta)).toEqual(['foo']);
+    expect(toArray(['foo'], meta)).toEqual(['foo']);
+    expect(toArray('', meta)).toEqual(['']);
+    expect(toArray(null, meta)).toEqual([null]);
```
gustavohenke: Do you see falsy values/nulls becoming an issue when they aren't transformed into `[]`?

gustavohenke: For clarification: I wonder if other users won't ask for a fine-grained control like in `optional`.

But nonetheless, I'm fine with whatever 👍

soulchild: I was wondering the same thing while implementing the test cases and didn't come to a conclusion on that. You're talking about transforming `[4,5,null,'',6]` to `[4,5,6]` by means of specifying some sort of `checkFalsy` and/or `checkNull` modifiers, right? I think we could always add options for that at a later time if the need arises. What do you think?

gustavohenke: > You're talking about transforming `[4,5,null,'',6]` to `[4,5,6]` by means of specifying some sort of `checkFalsy` and/or `checkNull` modifiers, right?

Not that, but whether `null` should become `[]` (like `undefined`), and then that could evolve even crazier... but oh well.

As I said, this code looks fine to me, so will merge it.

gustavohenke:
Hello @soulchild, sorry for the delay. This is now in [`v6.0.0`](https://github.com/express-validator/express-validator/releases/tag/v6.0.0) 🚢

PR #750, opened 2019-07-29
Update feature-custom-validators-sanitizers.md

alabobriggs:
The code for checking if password confirmation matches password does not behave expected as it always returns false even when it matches. Adding `.not()` before .custom() solves the issue.


alabobriggs:
> Why is this change fixing the problem?
> 
> Are you sure this isn't hiding any other bug with express-validator?
> Because a custom validator that returns `true` should pass, as documented.

When .not() is added before custom on that page it returns false. Try out that code on an existing code base and you'll see the error

alabobriggs:
@gustavohenke closing

Issue #742, opened 2019-07-18
"optional" with parameters "checkFalsy", "nullable" express-validator 6.1.1


GreyFoxRU:
Hello!
After upgrading to version 6.1.1, I ran into a problem: in ValidationSchema, the type for optional changed.
Previously used the design
```typescript
test: {
        optional: {
          options: {
            nullable: true,
            checkFalsy: true
           },
        },
        in: ['body'],
        isString: true
      }
```
but now the type is `optional ?: Partial <Optional> | true`, and it is not clear how to write now, because
```typescript
test: {
        optional: {
          nullable: true,
          checkFalsy: true
        },
        in: ['body'],
        isString: true
      }
```
it seems does not work. For example, I created a simple project ([test.zip](https://github.com/express-validator/express-validator/files/3405502/test.zip)), sending a request with `body.test = null`, server returns an error, as if there isnt setting `optional: { nullable: true }` at all.

The question of this has already been raised in previous versions: #608.

PR #765, opened 2019-08-19
Fix typing for optional

connorjayfitzgerald:
There's a bug in the typings when marking a field as optional.
For example, this flags as an error
```javascript
checkSchema({
  foo: {
    optional: {
      options: {
        nullable: true,
        checkFalsy: true,
      }
    }
  }
})
```

but allows the following:

```javascript
checkSchema({
  foo: {
    optional: {
      nullable: true,
      checkFalsy: true,
    }
  }
})
```

However, this has no effect.

This has been addressed previously in #693 but it looks like this has been regressed.

Also see #742

### DIFF
--- a/src/middlewares/schema.spec.ts
+++ b/src/middlewares/schema.spec.ts
@@ -146,7 +146,7 @@ describe('on each field', () => {
   });
 
   it('can be marked as optional', () => {
-    const chain = checkSchema({
+    const schema = checkSchema({
       foo: {
         optional: true,
       },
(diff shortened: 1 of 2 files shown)


src/middlewares/schema.spec.ts
```diff
@@ -146,16 +146,29 @@ describe('on each field', () => {
...
+      },
+    });
 
-    expect(chainToContext(chain).optional).toEqual({
+    expect(chainToContext(schema[0]).optional).toEqual({
       checkFalsy: false,
       nullable: false,
     });
+    
+    expect(chainToContext(schema[1]).optional).toEqual({
```
gustavohenke: I have pushed to your fork this change to the test, so it's a bit more useful!

gustavohenke:
Hello! This is out with [v6.2.0](https://github.com/express-validator/express-validator/releases/tag/v6.2.0)! 🚢

PR #776, opened 2019-09-19
Update mobile phone locale options (#772)

### LINKED ISSUE #772: Provide the 'pt-BR' mobile phone locale option
The [validator.js](https://github.com/validatorjs/validator.js) project already has the `pt-BR` mobile phone locale according to this file: [isMobilePhone.js](https://github.com/validatorjs/validator.js/blob/master/src/lib/isMobilePhone.js).

However, it is not possible to use this locale type since the express-validator library does not map this value in the MobilePhoneLocale options. I think that is necessary to update this file: [options.ts](https://github.com/express-validator/express-validator/blob/master/src/options.ts) 

Expected behavior:

```js
import { check } from 'express-validator';

check('attribute').isMobilePhone('pt-BR');
```

jeduardocosta:
This pull request updates the mobile phone locale by adding the `pt-BR' option, as the validator.js library already has this option for use.

Resolves #772

### DIFF
--- a/src/chain/validators-impl.spec.ts
+++ b/src/chain/validators-impl.spec.ts
@@ -148,25 +148,6 @@ describe('#isString()', () => {
   });
 });
 
-describe('#isMobilePhone()', () => {
-  it('checks if context is a valid mobile phone for pt-BR locale', async () => {
-    validators.isMobilePhone('pt-BR');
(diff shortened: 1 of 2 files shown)


gustavohenke:
Shipped in v6.3.0!

Issue #759, opened 2019-08-08
Documentation should update example for Typescript


TranBaVinhSon:
Documentation should update example for Typescript 👍 
https://express-validator.github.io/docs/

islasjuanp:
Hi @TranBaVinhSon . I agree with you. A Typescript example might be very useful for people to start using the library. I've just created a PR #788 with a little example. Please, let me know if you think that something is missing.

PR #788, opened 2019-10-16
docs: Add basic example written in Typescript

islasjuanp:
I've updated the example written in Typescript.  I think this doc might help other people to start using the library. 

Related to issue #759

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -31,6 +31,22 @@ app.post('/user', (req, res) => {
   }).then(user => res.json(user));
 });
 ```
+It can also be written in TypeScript: 
+```typescript
+import * as express from 'express';
+const app: express.Application = express();
+
+app.use(express.json());
+app.post(
+  '/user',
+  (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    User.create({
+      username: req.body.username,
+      password: req.body.password,
+    }).then(user => res.json(user));
+  }
+);
+```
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -57,6 +73,31 @@ app.post('/user', [
 });
 ```
 
+The validations can also be written in TypeScript in a similar way: 
+
+```typescript
+// ...rest of the initial code omitted for simplicity.
+import { check, validationResult } from 'express-validator/check';
```
fedeci, line 80:
```suggestion
import { check, validationResult } from 'express-validator';
```

fedeci (commented):
Hi @islasjuanp I don't think this should be closed as it may be useful for developers, but I would add something like a tabbed "bar" to select between TS and JS.

gustavohenke:
> I would add something like a tabbed "bar" to select between TS and JS.

Yes! This would be optimal.
Looks like it's available on Docusaurus now: https://v2.docusaurus.io/docs/markdown-features/#multi-language-support-code-blocks

fedeci:
Yes, if I am not wrong it is available also with `docusaurus@1.x`.

Should we update to `v2` or implement this using `v1`?

gustavohenke:
indeed is available on v1 :)
https://v1.docusaurus.io/docs/en/doc-markdown#language-specific-code-tabs

A move to v2 looks like more evolved work, so it can stay for later.

PR #821, opened 2020-01-19
matchedData: consider unused fields as optional data

### LINKED ISSUE #819: matchedData still contains data for a validation chain using .if() that returns false.
I'm not sure if this is a bug or if I'm not understanding the `onlyValidData` option for `matchedData()` correctly, but I have a validation setup like:

```
check('someObject.nestedObject.propertyA')
     .if((value, { req }) => req.body.someObject.nestedObject)
     .exists()
     ...
```

When I call `matchedData(req)`, I get back:
```
{ 
  ...,
  someObject: {
     nestedObject: {
          propertyA: undefined
     }
  },
  ...
}
```
though I would expect `matchedData` to **not** include the `propertyA` property. It’s not a huge deal as there are obvious ways to work with the output that’s currently being returned.

gustavohenke:
Adds an API in the context to mark fields as used, and use that in `matchedData` to treat unused ones as optional.

**This is a breaking change.** Chains that have no validators or sanitizers will no longer be included in the result of `matchedData()` by default.

Fixes #819

### DIFF
--- a/src/matched-data.spec.ts
+++ b/src/matched-data.spec.ts
@@ -62,6 +62,41 @@ it('does not include valid data from invalid oneOf() chain group', done => {
   });
 });
 
+describe('with chains that contain .if()', () => {
+  it('does not include data of chain with early failed condition', done => {
+    const req = {
+      query: { foo: 'foo' },
+    };
+
+    const middleware = check('foo')
+      .if((value: any) => /\d/.test(value))
+      .equals('foo');
+
+    middleware(req, {}, () => {
+      expect(matchedData(req)).toEqual({});
+      done();
+    });
+  });
+
+  it('includes data of chain with late failed condition', done => {
+    const req = {
+      query: { foo: 'foo' },
+    };
+
+    const middleware = check('foo')
+      .equals('foo')
+      .if((value: any) => /\d/.test(value))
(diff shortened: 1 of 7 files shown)


src/matched-data.spec.ts
```diff
@@ -108,7 +143,7 @@ describe('when option locations is defined', () => {
       query: { baz: true },
     };
 
-    check(['foo', 'bar', 'baz'])(req, {}, () => {
+    check(['foo', 'bar', 'baz']).exists()(req, {}, () => {
```
gustavohenke: How I identified the breaking change.

src/matched-data.spec.ts
```diff
@@ -62,6 +62,41 @@ it('does not include valid data from invalid oneOf() chain group', done => {
...
+  });
+
+  it('includes data of chain with late failed condition', done => {
+    const req = {
+      query: { foo: 'foo' },
+    };
+
+    const middleware = check('foo')
+      .equals('foo')
+      .if((value: any) => /\d/.test(value))
```
gustavohenke: This could use something that would actually work if used in production, `/\d/` will never pass as `foo` doesn't have any digits.

PR #840, opened 2020-02-26
Update feature-sanitization.md

vais:


vais:
Ooh, I see now - my bad! Maybe this:

> The `notifyOnReply` field isn't validated, but you can still call `toBoolean` sanitizer on the same `check` function to convert it to a JavaScript boolean

Or leave it as is, I was just a little confused here.

Thank you, @gustavohenke

PR #849, opened 2020-03-10
Bail support in schemas

### LINKED ISSUE #848: Usage of bail() in validation schema
Hello there!
Is there a way to get similar behaviour to what bait() method does using Schema Validation? The desired behaviour is to stop validation chain after a specific validation failed (`exists` in my case)

tommmn:
Added bail:boolean type to ValidatorSchemaOptions
Added isBailed flag and changed forEach function to some
Added flag to halt the validation chain in case of bail
Added tests to schema.spec.ts

Fixes #848


tommmn:
I forgot to implement some critic code. I will update as soon as I finish.

PR #864, opened 2020-04-28
correction in function title matchedData

philipe-vieira:
the function format was incorrect, in the form **`matchedData(req[, options])`**. Where the correct would be **`matchedData(req, { options })`**.


philipe-vieira:
I'm sorry, I really made a mess. I didn't know that the docs wrote like that. I thought it was just a small typo. again I'm sorry.

Issue #784, opened 2019-10-08
Async customSanitizer


ajmariduena:
Hi guys,

Im trying to use a promise on a `customSanitizer` with the following code:

```js
.customSanitizer(async (value) => {
    const User = await User.findOne({
        where: {
            id: value,
        },
    });

    return User;
})
```

But promise is not being resolved:
`{ user: Promise { <pending> } }`

Is there a way to do this?.

gustavohenke:
That's expected. There's no support for async sanitizers at the moment.
It shouldn't be hard to implement, though.

ilivss:
Damn I wasted like 4 hours searching for solution untill I found this post. Is there a workaround in the meantime?

NathanPB:
Is anyone already working on that?

fedeci:
Fixed in #919

PR #919, opened 2020-08-25
Made customSanitizer await for async functions

NathanPB:
I made this little tweak to make possible use customSanitize in async environments. I did not made a proper error handler because I don't know how to integrate that with the validators, so I don't think the work of this feature is done, that's just a draft. This is related to #784 

An example use case for this would be: ``body('user').customSanitizer(async(id) => await users.findById(id))``

Also made the tests for that

### DIFF
--- a/src/context-items/sanitization.ts
+++ b/src/context-items/sanitization.ts
@@ -13,8 +13,16 @@ export class Sanitization implements ContextItem {
   async run(context: Context, value: any, meta: Meta) {
     const { path, location } = meta;
 
+    const runCustomSanitizer = async () => {
+      const sanitizerValue = this.sanitizer(value, meta);
+      if (sanitizerValue instanceof Promise) {
+        return await sanitizerValue;
+      }
+      return sanitizerValue;
+    };
+
     const newValue = this.custom
-      ? (this.sanitizer as CustomSanitizer)(value, meta)
+      ? await runCustomSanitizer()
       : (this.sanitizer as StandardSanitizer)(toString(value), ...this.options);
 
     context.setData(path, newValue, location);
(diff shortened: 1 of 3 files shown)


src/context-items/sanitization.ts
```diff
@@ -13,8 +13,16 @@ export class Sanitization implements ContextItem {
   async run(context: Context, value: any, meta: Meta) {
     const { path, location } = meta;
 
+    const runCustomSanitizer = async () => {
+      const sanitizerValue = this.sanitizer(value, meta);
+      if (sanitizerValue instanceof Promise) {
+        return await sanitizerValue;
+      }
+      return sanitizerValue;
```
gustavohenke, line 21: I think this can simply be

```suggestion
      return Promise.resolve(sanitizerValue);
```

so that we can have the sync or async value wrapped in a promise, ready to be treated in case there's a rejection.

gustavohenke (commented):
Draft is already looking good to me!
What do you suggest that we do with the errors?

NathanPB:
> What do you suggest that we do with the errors?

I'd suggest automatically send a 500 http error code and abort the request at all.

Add the error to the error list given by ``validationResult(req)`` seems wrong because usually people just send a 400 when there are errors present, and a 400 error code would not be the case for errors thrown in the sanitizers (5xx instead)

gustavohenke (approved):
Cool! I think that an error will already cascade up to the route definition, so this LGTM as is.

gustavohenke:

wwei-flux:
Thanks for the feature. Should the document at https://express-validator.github.io/docs/custom-validators-sanitizers.html#custom-sanitizers be updated?

fedeci:
Sure @wwei-flux, you can raise a PR if you want!

PR #925, opened 2020-09-14
Add 'replace' sanitizer

### LINKED ISSUE #924: New replace sanitizer
Hello, what do you think of a new custom sanitizer that replaces the current value with a user chosen one if current value is one of those contained in a given array?

eg. to make it clear
```js
// 'name' is null
// the sanitizer will replace 'name' value with undefined if [null, undefined, NaN, ''] array includes 'name' value
check('name').replace(undefined, [null, undefined, NaN, ''])
```

This would help to maintain consistency in data types across the program.

fedeci:
This PR adds 'replace' sanitizer.
See more in the linked issue.
This closes #924

### DIFF
--- a/src/chain/sanitizers-impl.ts
+++ b/src/chain/sanitizers-impl.ts
@@ -8,9 +8,21 @@ import { Sanitizers } from './sanitizers';
 export class SanitizersImpl<Chain> implements Sanitizers<Chain> {
   constructor(private readonly builder: ContextBuilder, private readonly chain: Chain) {}
 
+  // custom sanitizers
   customSanitizer(sanitizer: CustomSanitizer) {
     this.builder.addItem(new Sanitization(sanitizer, true));
     return this.chain;
+  }
+  default(default_value: any) {
+    return this.customSanitizer(value =>
+      [undefined, null, NaN, ''].includes(value) ? default_value : value,
+    );
+  }
+  replace(values_to_replace: any, new_value: any) {
+    if (!Array.isArray(values_to_replace)) {
+      values_to_replace = [values_to_replace];
+    }
+    return this.customSanitizer(value => (values_to_replace.includes(value) ? new_value : value));
   }
 
   // Standard sanitizers
@@ -21,11 +33,6 @@ export class SanitizersImpl<Chain> implements Sanitizers<Chain> {
 
   blacklist(chars: string) {
     return this.addStandardSanitization(validator.blacklist, chars);
-  }
-  default(default_value: any) {
-    return this.customSanitizer(value =>
-      [undefined, null, NaN, ''].includes(value) ? default_value : value,
-    );
   }
   escape() {
     return this.addStandardSanitization(validator.escape);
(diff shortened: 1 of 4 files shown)


src/chain/sanitizers-impl.ts
```diff
@@ -34,6 +34,9 @@ export class SanitizersImpl<Chain> implements Sanitizers<Chain> {
   normalizeEmail(options?: Options.NormalizeEmailOptions) {
     return this.addStandardSanitization(validator.normalizeEmail, options);
   }
+  replace(new_value: any, values_to_replace: any[] = [null, undefined, NaN, '']) {
+    return this.customSanitizer(value => (values_to_replace.includes(value) ? new_value : value));
```
gustavohenke, line 38: uh... I was ready to complain about using `NaN` here, but seems like I would be mistaken:

```
gustavohenke ~ $ node
Welcome to Node.js v12.16.1.
Type ".help" for more information.
> [NaN].includes(NaN)
true
> NaN === NaN
false
```

src/chain/sanitizers.ts
```diff
@@ -10,6 +10,7 @@ export interface Sanitizers<Return> {
   ltrim(chars?: string): Return;
   normalizeEmail(options?: Options.NormalizeEmailOptions): Return;
   rtrim(chars?: string): Return;
+  replace(new_value: any, values_to_replace?: any[]): Return;
```
gustavohenke, line 13: I wonder if having `values_to_replace` be optional won't make it confusing, e.g.

```js
check('name').replace('John')
```

could be interpreted as "replace `name` value with `"John"`".

WDYT of swapping them?

fedeci: You mean to swap `new_value` and `values_to_replace`, right?
```typescript
replace(values_to_replace: any[], new_value: any): Return;
```
And what about if providing an empty array as `values_to_replace` would use default `[null, undefined, NaN, '']`.
```typescript
// declaration
function replace(values_to_replace: any[], new_value: any) {
  const v = values_to_replace.length !== 0 ? values_to_replace : [null, undefined, NaN, '']
  ...
}

// use case 
// would replace [null, undefined, NaN, ''] with undefined
check('name').replace([], undefined)
```

gustavohenke: > Another solution could be to remove the default array, and to add it as a different sanitizer called `default()` instead.

This sounds good to me! Are you willing to tackle it in this PR, or a separate one? Either way is fine.

fedeci: I think separating it may be a better choice.

src/chain/sanitizers.ts
```diff
@@ -10,6 +10,7 @@ export interface Sanitizers<Return> {
   ltrim(chars?: string): Return;
   normalizeEmail(options?: Options.NormalizeEmailOptions): Return;
   rtrim(chars?: string): Return;
+  replace(values_to_replace: any[], new_value: any): Return;
```
gustavohenke, line 13: Small nitpick:

```suggestion
  replace(values_to_replace: any, new_value: any): Return;
```

And then you can cast the value to an array if not already one, in order to optimise for one value replacements?

gustavohenke (commented):
Hey! Unfortunately it seems like merging the PR for `default` caused some conflicts on this one 😕

src/chain/sanitizers.ts
fedeci: Sure!

gustavohenke:

PR #926, opened 2020-09-17
docs: Add example sequential processing of the validation chain

checnev:
I have added the example to run validators sequentially in the documentation.
This is related to https://github.com/express-validator/express-validator/issues/638#issuecomment-530999622


gustavohenke (approved):
LGTM, thanks for sharing your tip!

PR #930, opened 2020-09-25
Add 'default' sanitizer

### LINKED ISSUE #856: Allow defaut values
Currently there is not an option to add a default value to an *optional* parameter. I tried implementing this with a CustomSanitizer and CustomValidator but it didn't work because they don't even run when the parameter is set to optional.  Here is an example:

```javascript
checkSchema({
  'params.myArray': {
    in: ['body'],
    isArray: true,
    optional: true, // if it's not set to optional the array validation fails
    customSanitizer: { options: (value) => [] }, //doesn't even run because the optional flag when the parameter doesn't exist
    errorMessage: 'should be an array'
  }
)
```

So I know I can implement another middleware or simply set the default value somewhere else, but I am validating a large object with `checkSchema` and I would save a lot of code by being able to do it. Issue #425 is related, but it's closed. It would be nice to have the option to set a default value or at least to sanitize a non existing value when optional is set

### LINKED ISSUE #929: New 'default' sanitizer
Adding this sanitizer would help to assign a default value if the input is one of `[null, undefined, NaN, '']`

eg.
```js
// if 'name' is one of [null, undefined, NaN, ''] it will be replaced with 'John'
check('name').default('John')
```

fedeci:
This PR adds the 'default' sanitizer.
See more in #929.
This closes #929 and closes #856

### DIFF
--- a/docs/api-sanitization-chain.md
+++ b/docs/api-sanitization-chain.md
@@ -45,6 +45,22 @@ app.get('/object/:id', param('id').customSanitizer((value, { req }) => {
 }), objectHandler)
 ```
 
+### `.default(default_value)`
+> *Returns:* the current sanitization chain instance
+
+Replaces the current value with a default one if the current value is included in `['', null, undefined, NaN]`.
+
+```js
+app.post('/', [body('username').default('foo')], (req, res, next) => {
+  // 'bar'     => 'bar'
+  // ''        => 'foo'
+  // undefined => 'foo'
+  // null      => 'foo'
+  // NaN       => 'foo'
+  console.log(req.body.username);
+});
+```
+
 ### `.run(req)`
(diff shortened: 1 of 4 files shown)


gustavohenke (commented):
Can you please also add some docs about it here?

https://github.com/express-validator/express-validator/blob/41bd5d974ecedf7f214213efc836a23418fadef9/docs/api-sanitization-chain.md#L46-L48

fedeci:
> Can you please also add some docs about it here?
> 
> 
> 
> https://github.com/express-validator/express-validator/blob/41bd5d974ecedf7f214213efc836a23418fadef9/docs/api-sanitization-chain.md#L46-L48

Sure

fedeci:
Done!

gustavohenke (approved):
Lovely, thanks!

gustavohenke:

PR #967, opened 2020-12-27
feat(isBoolean): implement 'strict' option

### LINKED ISSUE #904: isBoolean evaluate non-boolean as boolean
Used express-validator version 6.4.0  

I validate as below  
```
body('flag').isBoolean().withMessage('Flag property only accepts boolean')
```

value of flag not only accept boolean, but these value also detected as if valid boolean: "true", "false", ["true"], ["false", "true"]

I worked around it with:  
```
body('flag').custom(flag=> flag=== true || flag=== false).withMessage('Flag property only accepts boolean')
```

fedeci:
## Description

Fixes #904

## To-do list

### DIFF
--- a/src/chain/validators-impl.spec.ts
+++ b/src/chain/validators-impl.spec.ts
@@ -132,6 +132,34 @@ describe('#isAlpha()', () => {
   });
 });
 
+describe('strict #isBoolean()', () => {
+  it('adds validator to the context', () => {
+    const ret = validators.isBoolean({ strict: true });
+
+    expect(ret).toBe(chain);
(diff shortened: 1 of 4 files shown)


gustavohenke (approved):
👍 thanks!

mriedem:
Really late to this but was thinking about using it for something like a header or query parameter validation which are always strings, this wouldn't work for that though since it's strictly checking the value is true or false. Doing it with a string would be to do something like using `JSON.parse(value)` and handling that gracefully if it wasn't a boolean string (`"true"` or `"false"`).

fedeci:
You can use the non-strict version of `isBoolean` that checks for `"true"`, `"false"`, `"0"` or `"1"`.

PR #952, opened 2020-11-30
chore(deps): bump validator to 13.5.1

fedeci:
## Description

Update `validator` to 13.5.0
This allows to implement new validators and to add support to more locales.

## To-do list

- [x] I have added tests for what I changed.

- [x] This pull request is ready to merge.

PR #969, opened 2020-12-30
fix(IsAlphaOptions.ignore): convert 'string[]' to 'string'

fedeci:
## Description

This addresses a type bug introduced in #952.

https://github.com/express-validator/express-validator/pull/963#issuecomment-752722295

## To-do list

### DIFF
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -82,7 +82,9 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isAfter, date);
   }
   isAlpha(locale?: Options.AlphaLocale, options?: Options.IsAlphaOptions) {
-    return this.addStandardValidation(validator.isAlpha, locale, options);
+    const ignore = Array.isArray(options?.ignore) ? options?.ignore.join('') : options?.ignore;
+    delete options?.ignore;
+    return this.addStandardValidation(validator.isAlpha, locale, { ...options, ignore });
   }
   isAlphanumeric(locale?: Options.AlphanumericLocale) {
     return this.addStandardValidation(validator.isAlphanumeric, locale);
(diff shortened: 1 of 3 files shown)


src/options.ts
```diff
@@ -394,7 +394,7 @@ export interface ContainsOptions {
 }
 
 export interface IsAlphaOptions {
-  ignore?: string[] | RegExp;
+  ignore?: string | RegExp;
 }
```
fedeci, line 398: This is potentially a breaking change!

gustavohenke: Indeed. A way to make it not be breaking would be to test the type of `ignore` in the `isAlpha` implementation, WDYT?

fedeci: Yes, I was thinking about that.

src/options.ts
```diff
@@ -394,7 +394,7 @@ export interface ContainsOptions {
...
 
 export interface IsAlphaOptions {
-  ignore?: string | RegExp;
+  ignore?: string[] | RegExp;
```
gustavohenke:
```suggestion
  ignore?: string[] | string | RegExp;
```

should give maximum compatibility 👍

src/chain/validators-impl.ts
```diff
@@ -82,7 +82,9 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isAfter, date);
   }
   isAlpha(locale?: Options.AlphaLocale, options?: Options.IsAlphaOptions) {
-    return this.addStandardValidation(validator.isAlpha, locale, options);
+    const ignore = Array.isArray(options?.ignore) ? options?.ignore.join('') : options?.ignore;
+    delete options?.ignore;
```
gustavohenke, line 86: This line can be safely removed, since it'll be overridden below anyway

gustavohenke (approved):

gustavohenke:
Published as v6.9.1!

PR #971, opened 2021-01-02
feat: add 'rename' context-item

### LINKED ISSUE #785: [Feature Request] Ability to rename parameters
We're currently assessing if we want to use [@hapi/joi](https://www.npmjs.com/package/@hapi/joi) or express-validator. I'm in favor of this package because our app is in typescript, while the joi package is not, meaning we'd have to maintain the types in a separate file / package. 

The only thing that's missing in this package is the ability to rename a field for backwards compatibility, see joi's API here:

https://hapi.dev/family/joi/?v=16.1.7#objectrenamefrom-to-options

Any thoughts on adding this as a sanitizer? Or some plugin API? 

I'd be willing to open an MR for this!

fedeci:
## Description
This allows to rename a property.

```js
// req: { body: { name: 'foo' } } becomes 
// req: { body: { username: 'foo' } }
body('name').rename('username').isString()

// req: { body: { names: ['foo', 'bar'] } } throws
body('names.*').rename('username').isString()

// req: { body: { name: 'foo', surname: 'bar' } } throws
body(['name', 'surname']).rename('username')

// req: { body: { name: 'foo', surname: 'bar' } } throws
body('name').rename('surname')
```

Closes #785 

## To-do list

- [ ] Work on returned errors
- [ ] Wildcards?
- [ ] Dotted property names (e.g. `1.0`)

---

### DIFF
--- a/src/context-items/rename.spec.ts
+++ b/src/context-items/rename.spec.ts
@@ -1,0 +1,67 @@
+import { ContextBuilder } from '../context-builder';
+import { Meta } from '../base';
+import { Rename } from './rename';
+
+it('the new path is identical to the old one', () => {
+  const context = new ContextBuilder().build();
+  const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+
+  expect(new Rename('foo').run(context, 'value', meta)).resolves;
+});
+
+it('throws an error if trying to rename more than one field', async () => {
+  const context = new ContextBuilder().setFields(['foo', 'bar']).build();
+  const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+
+  await expect(new Rename('_foo').run(context, 'value', meta)).rejects.toThrow();
+});
+
+describe('throws an error if using wildcards', () => {
+  it('in context fields', async () => {
+    const context = new ContextBuilder().setFields(['foo.*']).build();
+    const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+
+    await expect(new Rename('_foo').run(context, 'value', meta)).rejects.toThrow();
+  });
+
+  it('in new path', async () => {
+    const context = new ContextBuilder().setFields(['foo']).build();
+    const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+
+    await expect(new Rename('foo.*').run(context, 'value', meta)).rejects.toThrow();
+  });
+});
+
+it('throws an error if the new path is already assigned to a property', async () => {
+  const context = new ContextBuilder().setFields(['foo']).build();
+  const meta: Meta = {
+    req: {
+      body: {
+        foo: 'value',
+        _foo: 'bar',
+      },
+    },
+    location: 'body',
+    path: 'foo',
+  };
+
+  await expect(new Rename('_foo').run(context, 'value', meta)).rejects.toThrow();
+});
+
+it('throws an error if trying to rename to an existing property', () => {
+  const context = new ContextBuilder().setFields(['foo']).build();
+  const meta: Meta = {
+    req: {
+      body: {
+        foo: 'value',
+      },
+    },
+    location: 'body',
+    path: 'foo',
+  };
+
+  expect(new Rename('_foo').run(context, 'value', meta)).resolves;
+  expect(meta.req.body).toEqual({
+    _foo: 'value',
+  });
+});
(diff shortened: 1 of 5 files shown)


src/context-items/rename.ts
```diff
@@ -0,0 +1,30 @@
...
+        throw new Error('Cannot rename multiple fields.');
+      }
+      const field = context.fields[0];
+      if (field.includes('*') || this.newPath.includes('*')) {
+        throw new Error('Cannot use rename() with wildcards.');
+      }
+
+      if (_.get(req[location], this.newPath) !== undefined) {
+        throw new Error(`Cannot rename to req.${location}.${path} as it already exists.`);
+      }
```
fedeci, line 23: These errors are thrown at runtime, should we use `context.addError()` even if it is not a validation error?

src/chain/context-runner-impl.ts
```diff
@@ -43,14 +44,19 @@ export class ContextRunnerImpl implements ContextRunner {
             path,
           });
 
-          // An instance is mutable, so if an item changed its value, there's no need to call getData again
-          const newValue = instance.value;
+          if (contextItem instanceof Rename) {
+            // change instance path
+            instance.path = contextItem.newPath;
```
fedeci, line 49: Is it correct to change `instance.path`?

src/context-items/rename.ts
```diff
@@ -0,0 +1,30 @@
...
+    const { req, location, path } = meta;
+
+    if (path !== this.newPath) {
+      if (context.fields.length !== 1) {
+        throw new Error('Cannot rename multiple fields.');
+      }
+
+      if (this.newPath.includes('*')) {
+        throw new Error('Cannot use rename() with wildcards.');
+      }
```
gustavohenke, line 19: WDYT of moving this to the constructor/`.rename()`?

Just thinking that failing on server start might be a better developer experience than on request.

src/context-items/rename.ts
```diff
@@ -0,0 +1,30 @@
...
+
+      if (_.get(req[location], this.newPath) !== undefined) {
+        throw new Error(`Cannot rename to req.${location}.${path} as it already exists.`);
```
gustavohenke, line 22: I'm guessing that a client might, due to a bug, send more information than required and litter logs with 5XX errors (as I think it'll probably flow all the way to `next()`).

Perhaps it's best to remove this?

src/context-items/rename.ts
```diff
@@ -0,0 +1,30 @@
...
+      }
+
+      _.set(req[location], this.newPath, value);
+      _.unset(req[location], path);
```
gustavohenke, line 26: These should obey the `dryRun` option

src/context-items/rename.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { ContextBuilder } from '../context-builder';
+import { Meta } from '../base';
+import { Rename } from './rename';
+
+it('the new path is identical to the old one', () => {
```
gustavohenke, line 5: nitpicking, but I usually like to make test names read nicely with the `it`.
For example:

```suggestion
it('does nothing when the new path is identical to the old one', () => {
```

src/context-items/rename.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  const context = new ContextBuilder().setFields(['foo']).build();
+  const meta: Meta = { req: {}, location: 'body', path: 'foo' };
+
+  expect(new Rename('foo').run(context, 'value', meta)).resolves;
```
gustavohenke, line 9:
```suggestion
  await expect(new Rename('foo').run(context, 'value', meta)).resolves;
```

though I'm afraid that `.resolves` might not be asserting anything 🤔

src/chain/context-runner-impl.ts
gustavohenke: I'll guess _no_ because of the below:
https://github.com/express-validator/express-validator/blob/6715a6a10963ef29f0eaf3c9e25463187afeab99/src/context.ts#L66-L70

but we can always change behaviour.

And side comment, I'd just find it a bit more classy if we could use context's APIs instead of checking for the specific implementation of `Rename` here.

gustavohenke (changes requested):
ohey only 1.5 years later

PR #972, opened 2021-01-02
feat(context-builder): add 'getProperties' method

fedeci:
## Description
This allows to get the properties of the builder.
It can be used to create a library based on the context that will be built .
_(e.g. swagger docs generation)_

## To-do list


gustavohenke (commented):
> I don't know if it is better to expose the properties in this way or to build a "local" context

Hmm I don't mind either way at the moment. The Context interface is quite small, so this looks reasonable for now.

I don't recall if the `.builder` property is exposed/documented anywhere... do you think it would make more sense to put it in https://github.com/express-validator/express-validator/blob/7f15d2d963574f34e0e312a96228b7b0693df1cf/src/chain/context-handler.ts#L5-L9 instead?

fedeci:
Yes, for example in `check` is exposed here. 
Maybe we should just document this so that the end user could create his own context with all his methods and not only access properties through the builder.
https://github.com/express-validator/express-validator/blob/7f15d2d963574f34e0e312a96228b7b0693df1cf/src/middlewares/check.ts#L32-L40

gustavohenke (changes requested):
Will LGTM if we can add more docs than just this 😬 
- API entry
- Possibly an entry under "Features", if it's necessary to showcase more of it

gustavohenke (commented):
...did you want to address the previous review?

fedeci:
Oops nope, asked review for the wrong PR😄

PR #975, opened 2021-01-06
fix(toString): no longer throwing with custom Object.toString

fedeci:
## Description
https://github.com/express-validator/express-validator/issues/791#issuecomment-752924430

```js
const value = {
  toString: 'something',
};

// returns '[object Object]' instead of throwing
const v = toString(value)

```
## To-do list

### DIFF
--- a/src/utils.spec.ts
+++ b/src/utils.spec.ts
@@ -10,7 +10,7 @@ describe('#toString', () => {
     expect(toString(value)).toEqual('foo');
   });
 
-  it('test custom object toString not "function"', () => {
+  it('calls Object.prototype.toString() when custom one is not function', () => {
     const value = {
       toString: 'something',
     };
--- a/src/utils.spec.ts
+++ b/src/utils.spec.ts
@@ -1,7 +1,7 @@
 import { toString } from './utils';
 
 describe('#toString', () => {
-  it('test custom object toString function', () => {
+  it('calls custom toString function', () => {
     const value = {
       toString() {
         return 'foo';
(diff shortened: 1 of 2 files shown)


src/utils.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { toString } from './utils';
+
+describe('#toString', () => {
+  it('test custom object toString function', () => {
```
gustavohenke, line 4:
```suggestion
  it('calls custom toString function', () => {
```

src/utils.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    const value = {
+      toString() {
+        return 'foo';
+      },
+    };
+    expect(toString(value)).toEqual('foo');
+  });
+
+  it('test custom object toString not "function"', () => {
```
gustavohenke, line 13: Not sure if this one is good, what do you think?

```suggestion
  it('calls Object.prototype.toString() when custom one is not function', () => {
```

src/utils.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    const value = {
+      toString: 'something',
+    };
+
+    expect(toString(value)).toEqual('[object Object]');
+  });
+
+  it('test missing object toString', () => {
```
gustavohenke, line 21:
```suggestion
  it('calls Object.prototype.toString() when passed object does not have a toString function', () => {
```

gustavohenke (approved):
Left some comments, but it looks good, thanks!

PR #976, opened 2021-01-06
feat(oneOf): run imperatively

### LINKED ISSUE #974: Running oneOf validations imperatively
So I want to run validations imperatively according to [docs](https://express-validator.github.io/docs/running-imperatively.html). It works fine till I have `oneOf` validator added, which, obviously, doesn't have a `run` method and the code below will fail:

```js
const validationRules = [
  oneOf([
    query('chunk')
      .isString()
      .notEmpty(),
    query('chunk')
      .isArray()
      .notEmpty()
  ])
];

    await Promise.all(validations.map(validation => validation.run(req)));

    if (errors.isEmpty()) {
      return next();
    }

    res.status(400).json({ errors: errors.array() });
  };
};

app.post('/api/get-chunks', validate(validationRules), async (req, res, next) => {
  // request is guaranteed to not have any validation errors.
 // ...
});
```

fedeci:
## Description
Fixes #974 

## To-do list


src/middlewares/one-of.spec.ts
```diff
@@ -159,3 +159,41 @@ describe('error message', () => {
...
+    };
+    const error = new Error();
+    const spy = jest.spyOn(ContextRunnerImpl.prototype, 'run').mockRejectedValue(error);
+
+    const middleware = oneOf([check('foo').isBoolean()]);
+    await expect(middleware.run(req)).rejects.toThrow(error);
+    spy.mockRestore();
+  });
+
+  it('successful validation', async () => {
```
gustavohenke, line 188: Can you just make this read more phrasal?
E.g. `sets no error in context when successful` or similar to how it's in other tests, like 
https://github.com/express-validator/express-validator/blob/7f15d2d963574f34e0e312a96228b7b0693df1cf/src/middlewares/one-of.spec.ts#L73

gustavohenke (approved):

PR #977, opened 2021-01-06
feat(checkSchema): run imperatively

### LINKED ISSUE #822: oneOf and checkSchema cannot easily be run imperatively.
We would like to make use of the the "Running validations imperatively" example [found here](https://express-validator.github.io/docs/running-imperatively.html), but are running into issues with `oneOf` and `checkSchema`. Simply stated the middleware functions returned by these two methods do not implement the [validation chain API](https://express-validator.github.io/docs/validation-chain-api.html#runreq), specifically the `run` method.

If we adapt the imperative example as follows there is a runtime exception because the `run` is not defined (this is just a contrived example):

```
    await Promise.all(validations.map(validation => validation.run(req)));

    if (errors.isEmpty()) {
      return next();
    }

    res.status(422).json({ errors: errors.array() });
  };
};

app.post('/api/create-user', validate([
  oneOf([
    body('email').isEmail(),
    body('email').equals('admin')
  ]),
  body('password').isLength({ min: 6 })
]), async (req, res, next) => {
  // request is guaranteed to not have any validation errors.
  const user = await User.create({ ... });
});
```

Looking at the source code, it is clear that the middleware function returned by `oneOf()` does not have the same properties as those returned by `check()`.

```
const { body, oneOf } = require('express-validator')

body('test', 'Testing').isString().run
[Function: bound run]

oneOf([body('a').isInt(), body('b').isString()]).run
undefined
```

`checkSchema` is slightly easer to work around since it returns an array of validation chains that will have a `run` method, though the imperative example given in the documentation fails to handle this case as well if you mix `checkSchema` with other validations. `validate(checkSchema({...}))` works fine, but not `validate([param('example')..., checkSchema({...})])`. You need something like the following:

```
    await Promise.all(validations.map(validation => {
      // Handle a checkSchema
      if (Array.isArray(validation)) {
        return validation.map(subvalidation => subvalidation.run(req))
      }

      // Handle standard validation chain
      if (typeof validation.run === 'function') {
        return validation.run(req)
      }

      // Handle oneOf
      return validation(req, null, (err) => { ... }) // Dummy next method to handle possible error.
    }))

    if (errors.isEmpty()) {
      return next();
    }

    res.status(422).json({ errors: errors.array() });
  };
};
```

I could not find any documentation stating that `oneOf` and `checkSchema` are not supported for running imperatively, at least per the example middleware for running validations imperatively. Can they be updated to include a `run` method? Or perhaps the middleware example for running validations imperatively could be updated to include handling for `oneOf` and `checkSchema` (when it is used in a mixed array of validation chains)?

fedeci:
## Description
Closes #822 

## To-do list

### DIFF
--- a/src/middlewares/schema.spec.ts
+++ b/src/middlewares/schema.spec.ts
@@ -263,3 +263,21 @@ describe('on schema that contains fields with bail methods', () => {
     expect(context.errors).toHaveLength(1);
   });
 });
+
+it('run checkSchema imperatively', async () => {
+  const req = {
+    body: { foo: 'foo' },
+  };
+  const schema = checkSchema({
+    foo: {
+      exists: true,
+      isLength: {
+        options: {
+          min: 5,
+        },
+      },
+    },
+  });
+
+  expect(schema.run(req)).resolves;
+});
--- a/src/middlewares/schema.spec.ts
+++ b/src/middlewares/schema.spec.ts
@@ -271,13 +271,9 @@ it('run checkSchema imperatively', async () => {
   const schema = checkSchema({
     foo: {
       exists: true,
-      isLength: {
-        options: {
-          min: 5,
-        },
-      },
+      isString: true,
     },
   });
 
-  expect(schema.run(req)).resolves;
+  return expect(schema.run(req)).resolves;
 });
--- a/src/middlewares/schema.ts
+++ b/src/middlewares/schema.ts
@@ -104,6 +104,7 @@ export function checkSchema(schema: Schema, defaultLocations: Location[] = valid
   const run = async (req: Request) => {
     return await Promise.all(chains.map(chain => chain.run(req)));
   };
+
   return Object.assign(chains, { run });
 }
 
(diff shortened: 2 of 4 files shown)


src/middlewares/schema.spec.ts
```diff
@@ -263,3 +263,21 @@ describe('on schema that contains fields with bail methods', () => {
...
+      exists: true,
+      isLength: {
+        options: {
+          min: 5,
+        },
+      },
+    },
+  });
+
+  expect(schema.run(req)).resolves;
```
gustavohenke, line 282: this needs an `await` or `return` -- otherwise it may fail out of band

src/middlewares/schema.ts
```diff
@@ -59,7 +59,7 @@ const validLocations: Location[] = ['body', 'cookies', 'headers', 'params', 'que
 const protectedNames = ['errorMessage', 'in'];
 
 export function checkSchema(schema: Schema, defaultLocations: Location[] = validLocations) {
```
gustavohenke, line 61: Can we provide a custom type here?
E.g.
```ts
ValidationChain[] & { run(req: Request): Promise<unknown> }
```

src/middlewares/schema.spec.ts
fedeci: I forgot the `return`🤦🏻‍♂️

gustavohenke (approved):

PR #978, opened 2021-01-14
feat: add 'isObject' validator

### LINKED ISSUE #882: an easy way to make property required or allow json property
I have tried using **checkSchema**, **check**, **body**, to see if there is a way to make property required but without checking the content of it but was unable to find how to do it. Is that even possible?

fedeci:
## Description

This PR implements `isObject` validator to validate objects.
`strict` option defaults to `true` so that arrays and null values does not pass validation, but it can be marked as `false` to allow them.

Closes #882 

## To-do list

### DIFF
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -58,6 +58,14 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     );
   }
 
+  isObject(options: { strict?: boolean } = { strict: true }) {
+    return this.custom(
+      value =>
+        typeof value === 'object' &&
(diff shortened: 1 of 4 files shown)


src/chain/validators-impl.ts
```diff
@@ -58,6 +58,14 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     );
   }
 
+  isObject(options: { strict?: boolean } = { strict: true }) {
+    return this.custom(
+      value =>
+        typeof value === 'object' &&
+        (options.strict ? value !== null && !Array.isArray(value) : true),
```
gustavohenke, line 65: nice 👍

gustavohenke (approved):

PR #997, opened 2021-02-26
chore: update issue templates

fedeci:
I removed these tags in #991 by mistake


fedeci:
Bad Github way to handle this! They should create the branch in my fork so that I have permissions to push more commits to it.

Issue #906, opened 2020-07-23
Implementing 'bail()' with Wildcard Validations


westcoastsuccess:
There was some brief discussion in [this issue](https://github.com/express-validator/express-validator/pull/500) (comments 3 and 4) about the ability to use 'bail()' with wildcard validations - I couldn't find any subsequent documentation, and my attempt to use it like this does not work as I intended it - I still get an array with a validation failure from each object in the array of objects I pass:

```
 body("documents.*")
    .optional()
     .isString().withMessage("Please try that again.").bail(),
```
    
Does this functionality actually exist, and I'm not implementing it correctly? If not, is it on the roadmap? Or is there an alternative approach?

liqiang372:
`bail` indeed works, you can try it with some other simple cases.
Can you elaborate on what payload you are trying to validate, and we can see how to validate that?

westcoastsuccess:
Thanks very much for responding, Qiang.

An example payload would be an array of objects, with the upload details associated with a previously uploaded document:

It hits the server as stringified JSON: 

```
req.body.newAccount.documentUploads:

documentUploads: ''[{"fileFullLocation":"someURL.png","fileRelativeLocation":"someOtherURL.png","fileExtension":"png","fileName":"testFile.png","contentType":"image/png","originalFileName":"originalFileName.png","fileDescription":"","fileUploadedBy":"someUser","fileSize":"1948","id":"someID"}]'
```

It next hits middleware (prior to validations), which parses it into an object using JSON.parse (this isn't documented anywhere I could find, but appears necessary, even though, for example, my app uses bodyParser i.e. `app.use(bodyParser.json());` ):

```
function extractJSON(req, res, next){
    if(req.body.newAccount.documentUploads){
        req.body.newAccount.documentUploads = JSON.parse(req.body.newAccount.documentUploads);
        return next();
    } else {
        return next();
    }
}
```

Then it hits the validators - here, for testing, I add "`.not()`" in front of "`.isString()`" to force an error at every object value (the wildcard invocation is "`.*.*`" because I want it to test each object in a possible array of objects, and then each value in that particular object - my attempts to invoke this via "`body("newAccount.documentUploads.*)`" did not work that way):

```
body("newAccount.documentUploads.*.*")
    .optional()   
    .not().isString().withMessage("It looks like there's an issue with your uploaded documents. Please try that again.").bail()
[...]

PR #1000, opened 2021-03-01
v7.0.0 features

### LINKED ISSUE #990: v7.0.0 Release Plan
This is a release plan of the next major release of `express-validator`.
## Potentially breaking bug fixes 
- [x] Passing an Array to `toString()` should not return just the first item #791, #755, #704 (PR: #1002)
- [ ] Improve `oneOf`:
  - [x] Error messages #956 (PR: #1022)
  - [ ] Wildcards #950

## New features
- [x] Pluggable validators #1077 (PR: #1079)
- [ ] Custom validators: 
  - [ ] Multiple error messages #951 
- [ ] Allow to bail out on the first element failing when using `bail()` with wildcards #906 

## Others
- [x] Remove deprecated APIs (filter, old-check, sanitize) (PR: #993)

I selected just some of the most recent and requested features/fixes, but this list is supposed to change so proposals are welcome!

/cc @gustavohenke Can you create a branch (`v7.0.0-features`?) on this repo to use as the base for PRs? When we merge it we should use `Rebase and merge` instead of `Squash and merge` so that we can track of those breaking commits better. It would definitely simplify development.

fedeci:
## Description
This closes #990 
This is the base branch used to develop v7.0.0 features.

### When merging this we should use `Rebase and Merge` because we want to track commits independently.

---
PR #1000 🥳


gustavohenke (commented):
I think we've gotten a fair number of features implemented here now, + multiple custom validators in checkSchema, which is in the base PR.

The few ones missing (and that we should give a try) are
- Wildcard-level bail (#906)
- Request-level bail (#1214)
- Type inference (#1207)
- ~Globstar support (#1137 and probably a few more)~

...and we can probably chuck a few more niceties nobody asked for, if we want.

Note that the bail ones are quite easy to implement, but might cause a perf hit as many things need to become serial instead of parallel.

@fedeci Thoughts on merging now and then working on the docs for releasing?

gustavohenke:
> Thoughts on merging now and then working on the docs for releasing?

I'm proceeding with this ☝️

PR #1002, opened 2021-03-02
feat: change `#toString()` behaviour for arrays

fedeci:
## To-do list


src/context-items/standard-validation.spec.ts
```diff
@@ -33,26 +33,27 @@ const createTest = (options: { returnValue: any; addsError: boolean }) => async
 
 it('calls the validator with the value as a string', async () => {
   await validation.run(context, false, meta);
-  expect(validator).toHaveBeenLastCalledWith('false');
+  expect(validator).toHaveBeenNthCalledWith(1, 'false');
```
gustavohenke, line 36: ❤️

gustavohenke (approved):
This is great. Thanks!

gustavohenke:
oops, just realised you didn't mark it as ready for review 😓

fedeci:
😁 I was still thinking about a few edge cases, but if they all pass I will just implement tests.

gustavohenke:
Any edge cases that you could think of? 🙂

fedeci:
Maybe wildcards and our custom validators for arrays.

gustavohenke:
taking this one over!

gustavohenke:
> Maybe wildcards and our custom validators for arrays.

To answer this:
- wildcards automatically expand and end up running one by one anyway, e.g. `foo.*` becomes `foo[0..N]`
- custom validators are, well, custom. They don't need `toString`.

src/context-items/standard-validation.ts
```diff
@@ -10,12 +10,18 @@ export class StandardValidation implements ContextItem {
...
+    // Deliberately not calling it `toString` in order to not override `Object.prototype.toString`.
+    private readonly stringify = toStringImpl,
   ) {}
 
   async run(context: Context, value: any, meta: Meta) {
-    const result = this.validator(toString(value), ...this.options);
-    if (this.negated ? result : !result) {
-      context.addError(this.message, value, meta);
-    }
+    const values = Array.isArray(value) ? value : [value];
```
yerycs, line 19: @fedeci This row is causing issue for some validation.
For example, isInt, IsIn validators are not validating correctly for empty array value.
```
const validator = checkSchema({ field: { isInt:{} }});
const req = { body: { field: [] } };
await validator.run(req);

```
For above code, errors shouldn't be empty. But errors become empty because of above row.
Any idea?

fedeci: Yes, it was meant to be like that. If you want to validate each item of the array you need to use
```js
const validator = checkSchema({ 'field.*': { isInt:{} }});
```

yerycs: No, `field` should be number. But there can be case that `field` in request is array. In that case, error should be triggered.

yerycs: I think it is bug. Right? @fedeci
Any idea to fix?

Issue #813, opened 2019-12-16
contextsSymbol prevents proper encapsulation of middleware


maxarndt:
I wrote an npm package to reuse my express validation code. 

This package allows me to inject error handling as a middleware into my express app using the `.use()` method.
The validation for arbitrary routes can be called as shown in the following code snippet of my router:

```
const SampleController = require('./controller')
const { check } = require('express-validator')
const { throwIfFails, validators } = require('@maxarndt/some-package')

module.exports = function (router) {
  router.post('/testOriginalValidators', throwIfFails(check('username').isEmail(), check('something').isAlphanumeric()), SampleController.testHandler)
  router.post('/testCustomValidator', throwIfFails(check('deviceId').custom(validators.deviceId)), SampleController.testHandler)
}
```

`@maxarndt/some-package` looks like the following:

```
const { validationResult } = require('express-validator')
const ValidationError = require('./ValidationError')
const formatError = require('./errorFormatter')

module.exports.validationErrorHandler = function (err, req, res, next) {
  if (err instanceof ValidationError) {
    const formattedError = formatError(err)
    return res.status(formattedError.statusCode).send(formattedError.error)
  }
  next(err)
}

module.exports.throwIfFails = function (...checks) {
  return [...checks, throwIfError]
}

function throwIfError (req, res, next) {
  try {
    validationResult(req).throw()
    return next()
  } catch (error) {
    throw new ValidationError('Validation error', error.errors)
  }
}
```

## Problem

While this works like a charm in my component tests in the package project it fails when I use this npm package in another application. 
The `validationResult(req)` function does not extract any errors because the `contextsSymbol` does not match. ([Link to the line in `validation-result.ts`](https://github.com/express-validator/express-validator/blob/d82af97c4efb0897ccb1c9c9227a46c32003aba6/src/validation-result.ts#L62))

My problem seems to be, that I require `express-validator@6.3.0` in my private package and additionally in the application which uses my own npm package.

I can avoid this error by exporting your `check` function and using this one. But I didn't want to wrap your whole module.

Is this an intended behavior? And if that's the case why is it necessary to use such a symbol?  
I would have expected the validation to work properly.

gustavohenke:
Hey!
That's definitely not an intended problem. I just never really thought of it being used like this.

I guess reverting back to string keys should make it.
In the mean time, you can try deduping the packages to fix the issue.

gustavohenke:
...well, I actually fixed it already. v6.3.1 is out.

[...]

PR #1014, opened 2021-03-30
chore: document why not using `Symbol` for `contextsKey`

fedeci:
## Description

We change the context key from  a `string` type, which does not guarantee unicity, to a [`Symbol`](https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Symbol).

## To-do list

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -48,7 +48,7 @@ export type ValidationError =
       nestedErrors?: unknown[];
     };
 
-// Not using simbols because of #813
+// Not using Symbol because of #813
 export const contextsKey = 'express-validator#contexts';
 
 export interface InternalRequest extends Request {
--- a/src/base.ts
+++ b/src/base.ts
@@ -48,7 +48,7 @@ export type ValidationError =
       nestedErrors?: unknown[];
     };
 
-export const contextsKey = 'express-validator#contexts';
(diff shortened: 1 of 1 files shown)


gustavohenke (changes requested):
Hah, it used to be like this in the past 😂 
#813 303b5eb927f9f173412b5e45097f5edd9810ecb1 https://github.com/express-validator/express-validator/releases/tag/v6.3.1

I guess we should add some comments to that line...

src/base.ts
```diff
@@ -48,7 +48,7 @@ export type ValidationError =
       nestedErrors?: unknown[];
     };
 
-export const contextsKey = 'express-validator#contexts';
+export const contextsKey = Symbol('contexts');
```
fedeci, line 51:
```suggestion
// Not using simbols because of #813
export const contextsKey = 'express-validator#contexts';
```

src/base.ts
```diff
@@ -48,6 +48,7 @@ export type ValidationError =
       nestedErrors?: unknown[];
     };
 
+// Not using simbols because of #813
```
fedeci, line 51:
```suggestion
// Not using Symbol because of #813
```
Oops Italian grammar🤦🏻‍♂️

gustavohenke (approved):

PR #1022, opened 2021-04-21
feat(oneOf): add new error types

### LINKED ISSUE #956: Bug: oneOf returns all the errors of a group
## Describe the bug
when using `oneOf` with groups of validation it returns all the errors of a group

## To Reproduce
1. use oneOf
```
oneOf([ 
 [
  body('name').equals('foo'),
  body('age').exists().isInt()
 ],
 [
  body('name').equals('bar'),
  body('work').exists().isString()
 ]
])
```
2. send a body with one wrong filed
```
{
"name":"foo",
"age":"f"
}
```

## Expected behavior
```
{
      "msg": "Invalid value(s)",
      "param": "_error",
      "nestedErrors": [
        {
          "value": "f",
          "msg": "Invalid value",
          "param": "age",
          "location": "body"
        }
     ]
}
```

## Current behavior
 ```
{
      "msg": "Invalid value(s)",
      "param": "_error",
      "nestedErrors": [
        {
          "value": "f",
          "msg": "Invalid value",
          "param": "age",
          "location": "body"
        },
        {
          "value": "foo",
          "msg": "Invalid value",
          "param": "name",
          "location": "body"
        },
        {
          "msg": "Invalid value",
          "param": "work",
          "location": "body"
        },
        {
          "msg": "Invalid value",
          "param": "work",
          "location": "body"
        }
      ]
    }
```

## Express-validator version:
 - Version: 6.6.0

### PR DESCRIPTION
**Pull request by fedeci**
## Description

This closes #956 
We allow the user to choose between different error types:
- `grouped`: new default behaviour that groups the errors based on the original branches.
- `legacy`: old behaviour that joins  the errors from different branches together
- `leastErroredOnly`: returns only the group with less errors

## To-do list

### DIFF
--- a/src/middlewares/one-of.spec.ts
+++ b/src/middlewares/one-of.spec.ts
@@ -140,7 +140,7 @@ describe('error message', () => {
 });
 
 describe('should let the user to choose between multiple error types', () => {
-  let errors: OneOfErrorType[] = ['grouped', 'legacy'];
+  let errors: OneOfErrorType[] = ['grouped', 'flat'];
   it.each(errors)(`%s error type`, async errorType => {
     const req: InternalRequest = {
       body: { foo: true },
--- a/src/middlewares/one-of.spec.ts
+++ b/src/middlewares/one-of.spec.ts
@@ -1,7 +1,7 @@
 import { InternalRequest, contextsKey } from '../base';
 import { ContextRunnerImpl } from '../chain/context-runner-impl';
 import { check } from './validation-chain-builders';
-import { oneOf } from './one-of';
+import { OneOfErrorType, OneOfOptions, oneOf } from './one-of';
 
 const getOneOfContext = (req: InternalRequest) => {
   const contexts = req[contextsKey] || [];
@@ -55,17 +55,7 @@ describe('with a list of chains', () => {
 
     oneOf([check('foo').isInt(), check('bar').isInt()])(req, {}, () => {
       const context = getOneOfContext(req);
-      expect(context.errors).toHaveLength(1);
-      expect(context.errors).toContainEqual(
-        expect.objectContaining({
-          param: '_error',
-          nestedErrors: expect.arrayContaining([
-            expect.objectContaining({ param: 'foo' }),
-            expect.objectContaining({ param: 'bar' }),
-          ]),
-        }),
-      );
-
+      expect(context.errors).toMatchSnapshot();
       done();
     });
   });
@@ -91,18 +81,7 @@ describe('with a list of chain groups', () => {
 
     oneOf([[check('foo').isInt(), check('bar').isInt()], check('baz').isAlpha()])(req, {}, () => {
       const context = getOneOfContext(req);
-      expect(context.errors).toHaveLength(1);
-      expect(context.errors).toContainEqual(
-        expect.objectContaining({
-          param: '_error',
-          nestedErrors: expect.arrayContaining([
-            expect.objectContaining({ param: 'foo' }),
-            expect.objectContaining({ param: 'bar' }),
-            expect.objectContaining({ param: 'baz' }),
-          ]),
-        }),
-      );
-
+      expect(context.errors).toMatchSnapshot();
       done();
     });
   });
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/middlewares/one-of.ts [hunk 1]
```diff
@@ -9,16 +9,19 @@ const dummyItem: ContextItem = { async run() {} };
 
 export type OneOfCustomMessageBuilder = (options: { req: Request }) => any;
 
+export type OneOfErrorType = 'grouped' | 'leastErroredOnly' | 'legacy';
```
fedeci, line 12: Not sure about `leastErroredOnly`, any suggestion?

gustavohenke: No better names yet...

src/middlewares/one-of.spec.ts
```diff
@@ -151,15 +130,57 @@ describe('error message', () => {
...
-      expect(context.errors[0]).toHaveProperty('msg', 'keep trying');
+      expect(context.errors).toMatchSnapshot();
       expect(message).toHaveBeenCalledWith({ req });
       done();
     });
   });
 });
 
+describe('should let the user to choose between multiple error types', () => {
+  // TODO: Can't use it.each because it doesn't support done() in TypeScript
```
gustavohenke, line 143: I realise you probably wanted to keep consistency in the file, but since #976, we can use promises instead of callbacks 😉

src/middlewares/one-of.ts
```diff
@@ -9,16 +9,19 @@ const dummyItem: ContextItem = { async run() {} };
...
+
+export interface OneOfOptions {
+  message?: OneOfCustomMessageBuilder | any;
```
gustavohenke, line 15: `| any` flattens the type to `any`, which is why there was 2 `oneOf` signatures.
Do you want to do something like `type OneOfOptions = { message: something } | { message: any }`?

gustavohenke (changes requested):

src/middlewares/one-of.ts
fedeci: That seems to still erase the type. Maybe the two signatures approach is the only way?

gustavohenke: Probably? That'd look good to me.

src/middlewares/one-of.ts
gustavohenke: I like the name `grouped` as it describes well the behaviour.
What do you think of `flat` instead of `legacy`? It sounds exactly like the opposite of `grouped`.

As for `leastErroredOnly`... still no better ideas 😢

gustavohenke (commented):
Can we add docs too?

src/middlewares/one-of.ts
fedeci: `flat` sounds nice to me.

src/middlewares/one-of.ts
```diff
@@ -9,16 +9,22 @@ const dummyItem: ContextItem = { async run() {} };
...
 export function oneOf(
   chains: (ValidationChain | ValidationChain[])[],
-  message?: OneOfCustomMessageBuilder,
+  options: { message: OneOfCustomMessageBuilder; errorType?: OneOfErrorType },
 ): Middleware & { run: (req: Request) => Promise<void> };
 export function oneOf(
   chains: (ValidationChain | ValidationChain[])[],
-  message?: any,
+  options: { message?: any; errorType?: OneOfErrorType },
 ): Middleware & { run: (req: Request) => Promise<void> };
```
fedeci, line 21: Hoping that in the future typescript will help us to avoid these awful signatures.

fedeci:
Merging this, we still have some time to update `leastErroredOnly` before the release

src/middlewares/one-of.spec.ts
```diff
@@ -151,15 +130,76 @@ describe('error message', () => {
...
-      expect(context.errors[0]).toHaveProperty('msg', 'keep trying');
+      expect(context.errors).toMatchSnapshot();
       expect(message).toHaveBeenCalledWith({ req });
       done();
     });
   });
 });
 
+describe('should let the user to choose between multiple error types', () => {
+  let errors: OneOfErrorType[] = ['grouped', 'flat'];
```
fedeci, line 143:
```suggestion
  const errors: OneOfErrorType[] = ['grouped', 'flat'];
```

PR #1029, opened 2021-05-02
docs: warn users about returning `falsy` values in `.if()` with async condition

### LINKED ISSUE #1028: Using .if() Asynchronously
How to use .if() asnycly? 
ex : `body('blah').if((value, { req }) => {
  return db.Model.findOne({ where: { id:7 } }).then(dat => dat.uuid === req.body.theUuid) // I need to call model for uuid
})`

It doesn't work

Or if I use async in .if() : 

`body('blah').if(async (value, {req}) => {
 return false // Just return false
})`

It returns true, yet

I need to call db model, so can somebody help me, please?

fedeci:
## Description
Closes #1028 
Changing this behaviour should be as simple as moving the await on the previous line, but would become a breaking change and I am not 100% sure if it would make sense.

## To-do list

### DIFF
--- a/docs/api-validation-chain.md
+++ b/docs/api-validation-chain.md
@@ -127,6 +127,8 @@ The condition can be either:
   If it returns truthy or a promise that resolves, the validation chain will continue
   running. If it returns falsy, a promise that rejects or if it throws, the validation chain will stop.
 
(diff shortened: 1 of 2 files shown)


gustavohenke (approved):
LGTM. I also don't like the idea of changing this behaviour.

PR #851, opened 2020-03-12
Support bail in schemas

tommmn:
Fix #848

PR #1040, opened 2021-05-21
Fix schema if fields

### LINKED ISSUE #834: ValidationSchema, object validation. Parent field nullable.
Hello!
Using a validation scheme, I am trying to validate an object. I need the parent field can to be undefined, and the nested field not. 
Maybe I'm doing it wrong, but I know for sure that this works with arrays.

For example for array:
```typescript
  const validationSchema: ValidationSchema = {
    'attachments': {
      in: ['body'],
      optional: {
        options: {
          nullable: true
        }
      },
      isArray: true
    },
    'attachments.*.id': {
      in: ['body'],
      isUUID: true,
      errorMessage: 'value should be proper attachment id'
    },
  };
```
This will work, if I do not send the attachment field or send the `attachments: undefined`, the check will succeed

Now suppose that there may be an object in the body that implements the interface
```typescript
interface ISomeObject {
  parentField?: {
    nestedField1: string;
    nestedField2: number;
  };
}

const someObjectShema: ValidationSchema = {
  'parentField': {
    in: ['body'],
    optional: {
      options: {
        nullable: true
      }
    }
  },
  'parentField.nestedField1': {
    in: ['body'],
    isString: true
  },
  'parentField.nestedField2': {
    in: ['body'],
    isNumeric: true
  },
}
```

Here, if I do not send the parent Field in the body, the validator swears at the absence of nested fields.

Help me please resolve the issue.

Janssena:
## Description

Closes #834
In the [documentation](https://express-validator.github.io/docs/validation-chain-api.html#ifcondition), `.if(condition)` takes a property `condition`. Using schema validation however, the script only looks for `options` properties. The resulting discrepancy between schema validation and the validation chain API is confusing. In general I think schema validation is under documented, as most validation chain functions do not work as documented using schema validation.

Nevertheless, the above PR brings schema validation closer to the state of the docs. I had to create an if statement as well, seeing as you guys likely don't want every field to allow both the use of `options` and `condition`.

## To-do list

### DIFF
--- a/src/chain/validators.ts
+++ b/src/chain/validators.ts
@@ -1,4 +1,4 @@
-import { CustomValidator, DynamicMessageCreator, ConditionValidator } from '../base';
+import { CustomValidator, DynamicMessageCreator } from '../base';
 import * as Options from '../options';
 
 export interface Validators<Return> {
@@ -9,7 +9,6 @@ export interface Validators<Return> {
 
   // custom validators
   custom(validator: CustomValidator): Return;
-  if(condition: ConditionValidator): Return;
   exists(options?: { checkFalsy?: boolean; checkNull?: boolean }): Return;
   isArray(options?: { min?: number; max?: number }): Return;
   isObject(options?: { strict?: boolean }): Return;
--- a/src/chain/validators.ts
+++ b/src/chain/validators.ts
@@ -1,4 +1,4 @@
-import { CustomValidator, DynamicMessageCreator } from '../base';
+import { CustomValidator, DynamicMessageCreator, StandardValidator } from '../base';
 import * as Options from '../options';
 
 export interface Validators<Return> {
@@ -9,6 +9,7 @@ export interface Validators<Return> {
 
   // custom validators
   custom(validator: CustomValidator): Return;
+  if(condition: StandardValidator): Return;
   exists(options?: { checkFalsy?: boolean; checkNull?: boolean }): Return;
   isArray(options?: { min?: number; max?: number }): Return;
   isObject(options?: { strict?: boolean }): Return;
--- a/src/chain/validators.ts
+++ b/src/chain/validators.ts
@@ -1,4 +1,4 @@
-import { CustomValidator, DynamicMessageCreator, StandardValidator } from '../base';
+import { CustomValidator, DynamicMessageCreator, ConditionValidator } from '../base';
 import * as Options from '../options';
 
 export interface Validators<Return> {
@@ -9,7 +9,7 @@ export interface Validators<Return> {
 
   // custom validators
   custom(validator: CustomValidator): Return;
-  if(condition: StandardValidator): Return;
+  if(condition: ConditionValidator): Return;
   exists(options?: { checkFalsy?: boolean; checkNull?: boolean }): Return;
   isArray(options?: { min?: number; max?: number }): Return;
   isObject(options?: { strict?: boolean }): Return;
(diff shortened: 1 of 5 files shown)


fedeci:
Thanks! I will look deeper into this later, but at a first glance it seems that tests are failing due to types issues, could you fix please?

gustavohenke (commented):
Hey. I think we're indeed lacking some proper support for `if` in `checkSchema`, but I'm not sure if this is the way to go.

I'd prefer it we did only `if: conditionFn` instead of `if: { condition: conditionFn }` -- it should look cleaner.

Then TypeScript support obviously needs to be added - it's clearly not there now, and a test like you added will be of much help to ensure it's supported.

fedeci:
We may do the same for custom validators.

Janssena:
I'd say that that would be the nicest solution. It also makes more sense for the custom validators & sanitizers.

fedeci:
Right @Janssena, we should implement it like we did for `#bail()` in #851, but following an implementation similar to `#withMessage()`.
https://github.com/express-validator/express-validator/blob/3bc20464921d3a6751c0719affbba2e23a996062/src/middlewares/schema.ts#L96-L100
Follow what is done in that PR and it all should be good. It goes without saying that is you encounter any problems you can ask here!

Janssena:
So I did some of the initial work here, but am not sure of how to proceed with this. I kind of understand how the schema validation works now, but I don't know that much of typescript and your preferred architecture to really make anything final. I'd suggest this gets converted to a story with some initial work done. I made the type, an initial test, but am getting stuck on the type declarations of ValidatorSchemaOptions. I guess since having the conditionFn directly in the if object there needs to be a lot of refactoring or hacky fixes (like the if statement catching the conditionFn before options that I added, no way suggesting you should keep that). I would say it'd be a welcome change to the schema validation since it is quite strange to have this `option` field be so hardcoded into the function of the schema validation.

src/chain/validators.ts
```diff
@@ -10,6 +10,7 @@ export interface Validators<Return> {
   // custom validators
   custom(validator: CustomValidator): Return;
   exists(options?: { checkFalsy?: boolean; checkNull?: boolean }): Return;
+  if(conditionFn: ConditionFn): Return;
```
fedeci, line 13: The same as above. `if` is not a validator, but a context handler, so is defined in that file.

src/middlewares/schema.ts
```diff
@@ -12,7 +12,8 @@ type ValidatorSchemaOptions<K extends keyof Validators<any>> =
       errorMessage?: DynamicMessageCreator | any;
       negated?: boolean;
       bail?: boolean;
-    };
+    }
+    | ConditionFn;
```
fedeci, line 16:
```suggestion
      if?: CustomValidator | ValidationChain;
    };
```

src/middlewares/schema.ts
```diff
@@ -83,6 +84,10 @@ export function checkSchema(
         // Using "!" because typescript doesn't know it isn't undefined.
         const methodCfg = config[method]!;
 
+        if (method === "if") {
+          return methodCfg ? chain : null;
+        }
+
```
fedeci, line 90: I would add this check after the options handling so that everything is ordered correctly. Note that we have to add it before the `not` check as it would be run before it.
It should also be replaced like this so that it is type checked by TS:
```ts
if (isValidatorOptions(method, methodCfg) && methodCfg.if) {
  chain.if(methodCfg);
}
```

fedeci (changes requested):
I tried to explain better what we have to do. For any questions please ask.

fedeci:
@Janssena are you still working on this?

Janssena:
@fedeci I have been focusing on other projects recently, so did not put much work on this PR. Not sure when I will have to time to work on this again.

fedeci:
No problem! Can I or someone else fork your PR and continue working on it? (Obviously giving credits to your commits)

Janssena:
@fedeci Of course! Sorry I have not been that active to actually solve these issues.

PR #1049, opened 2021-06-12
fix: prioritize `withMessage` errors over the ones thrown by `CustomValidator`

### LINKED ISSUE #1047: Validation with `custom` followed by `withMessage` always uses thrown message.
## Describe the bug

There seems to be a regression of #203 from the v2 days or the documentation for `withMessage` is incorrect.

If you try to override the default message of a custom validator using `withMessage`, it ignores the provided message.

According to [the documentation of `withMessage`](https://express-validator.github.io/docs/validation-chain-api.html#withmessagemessage):
> `.withMessage(message)`
> Sets the error message for the previous validator.
> This will have precedence over errors thrown by a custom validator.

And, according to [the Dynamic Messages guide](https://express-validator.github.io/docs/custom-error-messages.html#custom-validator-level):
> If you're using a custom validator, then it may very well throw or reject promises to indicate an invalid value.
> In these cases, the error gets reported with a message that's equal to what was thrown by the validator:

...which a colleague initially read as it will _always_ use the message from the error.  So, we're not sure which it is supposed to be and it's been so long since #203 that it's hard to say if the intended behavior had been altered since then.

But having `withMessage` override a default error message from a custom validator to provide better context is a handy tool to have in a validation library.

## To Reproduce

A partial reproduction:

```js
const { body, validationResult } = require('express-validator');

const customValidation = () => {
  throw new Error("default message");
};

router.post("/test", [
  body("someValue")
    .custom(customValidation).withMessage("custom message"),
  (req, res) => {
    try {
      validationResult(req).throw();
      res.send("OK");
    }
    catch (err) {
      // Expect to find "default message" in the output for `someValue`.
      console.log(err.mapped());
      res.status(400).send("FAIL");
    }
  }
]);

```

## Expected behavior

I would _like_ it to use the custom error message provided by `withMessage`
...or the documentation updated to be more clear about how custom validator error messages are intended to be presented.

## Current behavior

Always uses the thrown error message; checked in code for `src/context-items/custom-validation.js` and it indeed preferentially uses `err.message` or `err` over `this.message` when catching an exception from the custom validator.

## Express-validator version:

- Version: Discovered on 6.10.1, verified that it still affects 6.11.1

JHawkley:
## Description

Fixes #1047.
- Adjusted tests to account for set validation messages overriding thrown errors.
  - They were split into two `describe` blocks, one with and one without a message set.
- Adjusted `src/context-items/custom-validation.ts` to favor the set message over the thrown message.
- Tweaks "Custom Validator Level" section of `docs/feature-error-messages.md` to clarify this overriding behavior.

## To-do list

### DIFF
--- a/docs/feature-error-messages.md
+++ b/docs/feature-error-messages.md
@@ -65,6 +65,13 @@ app.post(
 );
 ```
 
+You may still use the [`.withMessage()` method](api-validation-chain.md#withmessagemessage) to
+override any message obtained from the custom validator with a custom message that is more appropriate
+to the route's context.
+
+This is especially useful if you wish to reuse a custom validator function across several routes
+throughout the application.
+
 ### Field Level
 
 Messages can be specified at the field level by using the second parameter of the
--- a/src/context-items/custom-validation.spec.ts
+++ b/src/context-items/custom-validation.spec.ts
@@ -17,8 +17,6 @@ beforeEach(() => {
   jest.spyOn(context, 'addError');
 
   validator = jest.fn();
-  validation = new CustomValidation(validator, false);
-  validation.message = undefined;
 });
 
 // There are 4 tests that are pretty much the same with just small variations among them.
@@ -34,6 +32,11 @@ const createSyncTest = (options: { returnValue: any; addsError: boolean }) => as
 };
 
 describe('when not negated', () => {
+  beforeEach(() => {
+    validation = new CustomValidation(validator, false);
+    validation.message = 'nope';
+  });
+
   it(
     'adds error if validator returns falsy',
     createSyncTest({ returnValue: false, addsError: true }),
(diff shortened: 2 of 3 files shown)


JHawkley:
Oh, I should point out that this _could_ be a breaking change in some circumstances, even though it follows what the docs say about `withMessage`.

A developer _could_ have relied on the previous behavior with a specially constructed custom validator function:

```js
const dynamicValidator = (value) => {
  if (typeof value === "string") return Promise.reject("for string");
  if (typeof value === "number") return Promise.reject("for number");
  return Promise.reject(undefined);
};

// The custom message would only be used if the validator rejects with `undefined`
// using the previous behavior.
chain("someValue)
  .custom(dynamicValidator).withMessage("for something else");
```

It's unlikely, but I thought I should point it out.

fedeci (approved):
Thanks!
Since it is not a major breaking change I would release this in a minor, maybe with a special warning about it.

gustavohenke (approved):
❤️ nice one!

gustavohenke:
Published in v6.12.0!

PR #1061, opened 2021-07-14
Use readonly arrays for validator options

### LINKED ISSUE #1059: Argument of type 'readonly number[]' is not assignable to parameter of type 'any[]'
## Describe the bug

When I use:
`check("foo").isIn(myArray)`
where myArray is readonly array then I got an error:

> Argument of type 'readonly number[]' is not assignable to parameter of type 'any[]'.
>   The type 'readonly number[]' is 'readonly' and cannot be assigned to the mutable type 'any[]'.ts(2345)

## Expected behavior

isIn shouldn't report an error for readonly arrays.

## Express-validator version:

- Version: 6.12.0

gustavohenke:
## Description

Changes validators options from `type[]` to `readonly type[]` - we don't do any mutation to them, so there's no need to require a mutable array.

Example of the issue: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhgJwQLhggpnAJuANgT3jHwG0BdGAXhhIEYAaGAJkYGYyBuAKADMBXMMCgBLcDBwBlEAFt0UABbCwAcwAUSgA58oqOMXIBKGAG8Avl0ky5ilasQIDHIA

Fixes #1059

## To-do list

- [ ] I have added tests for what I changed


gustavohenke:
Looks like travis-ci.org has been halted, I'm trying to set us up on travis-ci.com now... 😕

gustavohenke:
(if the above doesn't work, we might have to use github actions...)

fedeci:
I would directly go eith GH actions that are better integrated.

fedeci (approved):

PR #579, opened 2018-05-13
Docs overhaul

gustavohenke:
**Coming soon next to you, a website with waaay better docs!**

screen shot 2018-05-14 at 00 47 38

This is using Facebook's [Docusaurus](https://docusaurus.io). It's 💯 awesome.

To-do list:
- [x] Migrate existing docs to new format
- [x] Build docs on CI
- [x] Publish them to gh-pages
- [x] Have a stupid basic logo
  - [x] Use it on the docs header
  - [x] Use it on the favicon
- [x] Prepare for versioning
- [x] Have "Edit on GitHub" buttons

Closes #515

PR #1066, opened 2021-07-15
Add basic github actions workflow

gustavohenke:
Adding a basic GitHub workflow with testing, linting, and docs building.
Testing happens on Node 12, 14 and 16, whereas linting and docs building is only on Node 16.

They should all run in parallel.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - name: Build docs
+        run: npm run docs:build
```
fedeci, line 53: Do we build docs just to check that nothing fails?

gustavohenke: Pretty much, I think.

I don't recall why I added it there back in [`f9aec48` (#579)](https://github.com/express-validator/express-validator/pull/579/commits/f9aec483c4944dbc1c8218fb5766ba886b377542).

fedeci (approved):

fedeci:
Maybe we have to also remove the old travis config file in this PR.

fedeci:
It is running everything twice because of the double `push` and `pull_request` listeners🤔

gustavohenke:
Yes 😃

fedeci:
I'm not sure hot to avoid it :)

gustavohenke:
It runs using the PR's latest commit and what would happen if it got merged onto the target branch.

It's a good thing - theoretically one could introduce a breaking change to `master`, and this PR be passing without nobody noticing until it's merged.

fedeci:
Isn't it the default behaviour for both the hooks? I think that it is simply calling both of them.

Issue #1010, opened 2021-03-16
buildCheckFunction default added only to req.body


shizpi:
## Describe the bug

Default value is always added to `req.body` when using buildCheckFunction to validate body and query.

## To Reproduce

```
const checkBodyAndQuery = buildCheckFunction(['body', 'query']);

export const validateInvoicesParams = [
  checkBodyAndQuery('id').optional().isNumeric(),
  checkBodyAndQuery('amount').optional().custom(isRangeArray),
  checkBodyAndQuery('starts-at').default(DEFAULT_STARTS_AT).toDate().isISO8601(),
  checkBodyAndQuery('ends-at').default(DEFAULT_ENDS_AT).toDate().isISO8601(),
];
```

## Expected behavior

Should add the default value for any missing param to both places instead of adding default only to `req.body`

## Current behavior

Adds default value only to `req.body`

## Express-validator version:

- Version: 6.10.0

fedeci:
This seems not to be caused by the `default()` sanitizer, but by `buildCheckFunction()` that evaluates field results as an union and not as an intersection. When the context is run if at least one of the specified locations passes the validation then the whole `check` passes.
You can reproduce the "bug" also running:
```js
checkBodyAndQuery('id').isInt(),
```
and performing a request with empty query params and `{ id: 1 }` as body.
As you can see it passes validation. 
To avoid this, you should specify different validation chains:
```js
[
  body('id').isInt(), 
  query('id').isInt(),
]
```

shizpi:
@fedeci Yes, I want it to pass the validation. It should be in query OR in body so the validation works as intended.

But when `starts-at` is missing in both places the default value is added only to req.body, instead of adding it also to `req.query`.

If I would validate explicitly `query('starts-at').default(DEFAULT_STARTS_AT).toDate().isISO8601(),` then the default would be added to `req.query`.

PR #1070, opened 2021-07-29
fix: check all locations if in is not specified.

nikli2009:
## Description

The `default` option doesn't work as expected when the field is in `req.query` and `in` is not specified.

## To-do list


src/context.ts
```diff
@@ -43,12 +43,12 @@ export class Context {
...
 
         // #331 - When multiple locations are involved, all of them must pass the validation.
-        // If none of the locations contain the field, we at least include one for error reporting.
+        // If none of the locations contain the field, we should check all locations since we are not sure which location the consumer uses.
         // #458, #531 - Wildcards are an exception though: they may yield 0..* instances with different
         // paths, so we may want to skip this filtering.
         if (instances.length > 1 && locations.length > 1 && !group.includes('*')) {
           const withValue = instances.filter(instance => instance.value !== undefined);
-          return withValue.length ? withValue : [instances[0]];
+          return withValue.length ? withValue : instances;
```
fedeci, line 51: This imply that even locations where `!Object.prototype.hasOwnProperty(location, param)` get the default value added🤔

nikli2009: Hi buddy, thanks for the review. :D 
Just for clarification, the line I changed (Line 51) used to make the `default` only works for `req.body` (which is the first item of `instances`), and I thought maybe it would probably makes more sense to make it works for all `location`s when consumer hasn't set any values for `in`.

So in the commit, I changed to `instances` and this makes the `default` feature in `checkSchema` to be more consistent with the other scenario that this lib currently supports already. [here](https://express-validator.github.io/docs/sanitization-chain-api.html#defaultdefault_value)

If we set up like 
```
app.use('/user', query('greeting').default('hiya'), userController);
```
and send request `GET /user`, the `req.query` will be transformed into `{ greeting: 'hiya' }`

So in short, like you commented, `!Object.prototype.hasOwnProperty(location, param)` does get the default value added... and this behavior is consistent with the other API [here](https://express-validator.github.io/docs/sanitization-chain-api.html#defaultdefault_value)

fedeci: Well #1070 is actually not related to this PR as written in the issue.
However the first problem that comes to my mind is that line 51 is not only used by `.default` and is shared across all the other validators and sanitizers. The fix may be quite though sadly :(

nikli2009: Got you, thanks for maintain the lib ! Glad to add more tests if can help save some verification effort (maybe another PR).

src/middlewares/schema.spec.ts
```diff
@@ -183,12 +183,29 @@ describe('on each field', () => {
...
+    expect(context.getData().every(instance => instance.value === 10)).toBe(true);
+    expect(req.query.undef).toBe(10);
+  });
 });
 
 describe('on schema that contains fields with bail methods', () => {
   it('stops validation chain with only one error', async () => {
     const schema = checkSchema({
       foo: {
+        in: ['params'],
```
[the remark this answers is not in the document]
nikli2009: Just update the tests and one thing I noticed is: 
When `in` is not specified then check all locations by default, user will be getting 5 errors from 5 default locations. However this means if use `exists` + `bail` method and the field is not in any location of the req object,  user will still get 5 errors.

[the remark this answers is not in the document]
fedeci: Was it like that before?

fedeci:
Hey @nikli2009 can you add a few more tests to check if #1010 is fixed by this?

nikli2009:
> Hey @nikli2009 can you add a few more tests to check if #1010 is fixed by this?

hey man, gladly ! checking today

src/middlewares/schema.spec.ts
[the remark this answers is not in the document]
nikli2009: Just updated the tests along with some explanation comments, it used to only generate `1` error after checking `req.body`, however as location is not specified, we can't actually know which location user intend to check.

nikli2009:
> Hey @nikli2009 can you add a few more tests to check if #1010 is fixed by this?

Test added, could you help review and let me know if need to add more

src/context.ts
```diff
@@ -43,12 +43,12 @@ export class Context {
         const locations = _.uniqBy(instances, 'location');
 
         // #331 - When multiple locations are involved, all of them must pass the validation.
-        // If none of the locations contain the field, we at least include one for error reporting.
+        // If `in` option is not specified, try to check all locations by default.
```
fedeci, line 46:
```suggestion
        // checkSchema: If `in` option is not specified, try to check all locations by default.
```

src/middlewares/schema.spec.ts
```diff
@@ -200,7 +233,57 @@ describe('on schema that contains fields with bail methods', () => {
...
+          },
+        },
+      },
+    });
+
+    const { context } = await schema[0].run({ query: {} });
+    // when location is not specified, checking each location one by one,
+    // if the expected field does not show up at all in any location, then there will be 5 errors.
+    // finally bails.
+    expect(context.errors).toHaveLength(allLocations);
```
fedeci, line 260: nit: we can hardcode `allLocations` in the tests since locations will unlikely change in length.

src/middlewares/schema.spec.ts
```diff
@@ -183,12 +183,45 @@ describe('on each field', () => {
...
+          options: 10,
+        },
+      },
+    })[0];
+
+    const req = { query: { undef: undefined } };
+    const { context } = await chain.run(req);
+
+    expect(context.getData().every(instance => instance.value === 10)).toBe(true);
+    expect(req.query.undef).toBe(10);
```
fedeci, line 200: Can we check that also `req.body.undef`, ... have value of `10`. We will fix this testing problem with fixtures soon :)

fedeci (approved):
Some nitty comments but look fine. Thanks!
/cc @gustavohenke patch or minor? It actually corrects a bug(?) but may break something🤔

src/middlewares/validation-chain-builders.spec.ts
```diff
@@ -48,6 +48,32 @@ describe('buildCheckFunction()', () => {
...
   });
+
+  describe('when at least 1 locations specified', () => {
+    // for #1010
+    it('should apply default value to each specified location if default option is not empty', async () => {
+      const custom = buildCheckFunction(['cookies', 'headers']);
+      const chain = custom('foo').default(10);
+
+      // reset req
+      req = {
```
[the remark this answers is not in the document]
nikli2009: good one! will update

src/middlewares/schema.spec.ts
nikli2009: sure 👌

nikli2009:
Tests updated + comment squashed 🥪🥪 .

gustavohenke (changes requested):
This is a breaking change.
For everybody using `check()` instead of the specialised functions, this can be a bit wild...

Test code:
```js
const req = {};
check('foo').isNumeric().run(req).then(() => {
  console.log(validationResult(req).array());
});
```

Output:

```js
[
  {
    value: undefined,
    msg: 'Invalid value',
    param: 'foo',
    location: 'body'
  },
  {
    value: undefined,
    msg: 'Invalid value',
    param: 'foo',
    location: 'cookies'
  },
  {
    value: undefined,
    msg: 'Invalid value',
    param: 'foo',
    location: 'headers'
  },
  {
    value: undefined,
    msg: 'Invalid value',
    param: 'foo',
    location: 'params'
  },
  {
    value: undefined,
    msg: 'Invalid value',
    param: 'foo',
    location: 'query'
  }
]
```

---

Looking at #1071, it's conceptually wrong that you don't pass `in` to the schema, and expect a certain request location to have the value you wanted.
You should probably use `matchedData()` instead.

But then, a few things I'm wondering:
- would it makes sense to differentiate between "all/multiple locations" vs "any location", so to appease to those that want the convenience of `check`?
- should we verify if the user actually tried to pass the validated field somewhere -- e.g. `/some?query` vs `/some` -- and prefer handling `req.query` instead of `req.body`?

PR #1072, opened 2021-07-31
fix: add type for `checkSchema` function return

nikli2009:
## Background

image
When getting a list of result by calling `checkSchema().run`, I was unable to use those handy methods on result as it was marked as `unknown` currently. 
Would be def great if we could add type support for the return of `checkSchema`, thanks.

## Description

Type definition for `checkSchema` function return.

## To-do list

### DIFF
--- a/src/middlewares/schema.ts
+++ b/src/middlewares/schema.ts
@@ -3,8 +3,8 @@ import { Validators } from '../chain/validators';
 import { DynamicMessageCreator, Location, Request } from '../base';
 import { ValidationChain, ValidatorsImpl } from '../chain';
 import { Optional, ReadonlyContext } from '../context';
-import { check } from './check';
 import { Result } from '../validation-result';
+import { check } from './check';
 
 type ValidatorSchemaOptions<K extends keyof Validators<any>> =
   | true
--- a/src/middlewares/schema.ts
+++ b/src/middlewares/schema.ts
@@ -2,8 +2,9 @@ import { Sanitizers } from '../chain/sanitizers';
 import { Validators } from '../chain/validators';
 import { DynamicMessageCreator, Location, Request } from '../base';
 import { ValidationChain, ValidatorsImpl } from '../chain';
-import { Optional } from '../context';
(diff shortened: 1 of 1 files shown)


fedeci (approved):
Thanks!🚀

src/middlewares/schema.ts
```diff
@@ -62,7 +63,7 @@ export function checkSchema(
   schema: Schema,
   defaultLocations: Location[] = validLocations,
 ): ValidationChain[] & {
-  run: (req: Request) => Promise<unknown[]>;
+  run: (req: Request) => Promise<(Result & { context: ReadonlyContext })[]>;
```
fedeci, line 66:
```suggestion
  run: (req: Request) => Promise<ResultWithContext[]>;
```

src/middlewares/schema.ts
```diff
@@ -2,7 +2,8 @@ import { Sanitizers } from '../chain/sanitizers';
 import { Validators } from '../chain/validators';
 import { DynamicMessageCreator, Location, Request } from '../base';
 import { ValidationChain, ValidatorsImpl } from '../chain';
-import { Optional } from '../context';
+import { Optional, ReadonlyContext } from '../context';
+import { Result } from '../validation-result';
```
fedeci, line 6:
```suggestion
import { ResultWithContext } from '../chain/context-runner-impl';
```

src/middlewares/schema.ts
```diff
@@ -2,7 +2,7 @@ import { Sanitizers } from '../chain/sanitizers';
 import { Validators } from '../chain/validators';
 import { DynamicMessageCreator, Location, Request } from '../base';
 import { ValidationChain, ValidatorsImpl } from '../chain';
-import { Optional } from '../context';
+import { ResultWithContext } from '../chain/context-runner-impl';
```
fedeci, line 5:
```suggestion
import { Optional } from '../context';
import { ResultWithContext } from '../chain/context-runner-impl';
```

fedeci:
Thanks again!

PR #1076, opened 2021-08-08
feat: add failing validator name to context errors

### LINKED ISSUE #1030: How to get failed error type
Hi i'm trying to get the error type failed to validate:

This is what o have when validation fail but i want to know what kind of fail was. like: required or maxlength etc.
{
    value: undefined,
    msg: 'Invalid value',
    param: 'lname',
    location: 'body'
  }

I need something like this:

{
    value: undefined,
    msg: 'Invalid value',
    param: 'lname',
    location: 'body',
    king: 'required'
  }

fedeci:
## Description
Closes #1030 

## To-do list

### DIFF
--- a/src/middlewares/schema.spec.ts
+++ b/src/middlewares/schema.spec.ts
@@ -113,6 +113,22 @@ describe('on each field', () => {
     );
   });
 
+  it('halts chain execution if "if" statement resolves to false', async () => {
+    const schema = checkSchema({
+      foo: {
+        isEmpty: {
+          if: (value: any) => {
+            return value !== '';
+          },
+        },
+      },
+    });
+
+    const { context } = await schema[0].run({ query: { foo: '' } });
+
+    expect(context.errors).toHaveLength(0);
+  });
+
   it('sets error message', () => {
     const chain = checkSchema({
       foo: {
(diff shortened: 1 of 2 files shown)


src/context-items/custom-validation.ts
```diff
@@ -17,14 +17,19 @@ export class CustomValidation implements ContextItem {
       // A promise that was resolved only adds an error if negated.
       // Otherwise it always suceeds
       if ((!isPromise && failed) || (isPromise && this.negated)) {
-        context.addError(this.message, value, meta);
+        context.addError(this.message, value, meta, 'customValidator');
```
gustavohenke, line 20: How satisfied are you with `customValidator` here? 🤔

gustavohenke (commented):
This will be a breaking change - which to avoid we should implement behind a setting.

Do you think there are any use cases other than making the default error message more useful?

I'm worried of introducing a way for hackers to tell exactly which implementation can be exploited to get through any HTTP call.

Also, extra download weight that may not be necessary to many.

src/context-items/custom-validation.ts
fedeci: No idea of how to improve🤔, maybe `.name` fallbacking to `'customValidator'`?

fedeci:
Why a breaking change? We are not radically changing an existing feature, but adding a new one.

Probably making it opt-in as part of `validationResult` custom formatters is better.

Issue #1077, opened 2021-08-08
Allow simple extension of custom validators/sanitizers for better readability and reusability


ysfaran:
## Is your feature request related to a problem? Please describe.

When I add a custom validator, which I need to reuse, I usually create a function like this:

```js
export const isToken = (value) => {
  if (typeof value !== "string") {
    throw Error("Value must be a string");
  }

  if (!value.startsWith("0x")) {
    throw Error("Value  must start with '0x'");
  }

  if (value.length !== 42) {
    throw Error("Value must be 42 characters long");
  }

  return true;
};
```

Usage: 

```js
body("token").custom(isToken),
```

I use the snippet `custom(isToken)` above for a lot of validations and also reuse it for different middlewars like `param` instead of `body`. This unfortunatley breaks the cleanness of the fluent api and makes it harder to read. You also need to import `isToken` everywhere you use it. For `customSanitizer` it's quite similar.

## Describe the solution you'd like

`express-validator` should allow to extend it's validators/sanitizers in an easier way, similiar to how `jest` does it with [`expect.extend`](https://jestjs.io/docs/expect#expectextendmatchers).

A good solution would be to export new functions from the library:

```js
import { extend } from "express-validator";
```

which will allow to add custom validators and sanitizers:

```js
extend.validators({
  isToken,
});

extend.sanitizers({
  toSomeSanitizer: /*..*/,
});
```

Usage afterwards:

```js
body("token").isToken().toSomeSanitizer(),
```

With this solution the API would look much cleaner and it would simplify reusability in contrast to `custom` or `customSanitizer`.

## Implementation Idea

The feature could be implemented by extending the `prototype` of https://github.com/express-validator/express-validator/blob/690cd636df27cd5d32ad4f0e3f6ca783d86f873f/src/chain/validators-impl.ts#L8

```js
export const extend = {
  validators(customValidators){
    for([name, validator] of Object.entries(customValidators)){
      // note: currently this.custom doesn't seem to support options
      ValidatorsImpl.prototype[name] = function(options) { return this.custom(validator, options); }
    }
  }
}
  
  // sanitizers() ...
}
```
It should also be supported to pass options to custom validators, e.g.:

```js
export const isToken = (value, { length = 42 } = {} ) => {
  // ...

  if (value.length !== length) {
    throw Error(`Value must be ${length} characters long`);
  }

  return true;
};

body("token").isToken({ length: 18 }),
```

[...]

PR #1079, opened 2021-08-11
feat(extend): allow adding of custom validators globally (#1077)

ysfaran:
## Description

Everything is already described in https://github.com/express-validator/express-validator/issues/1077.
This PR is my suggested implementation for the validators part. I didn't implement it for sanitizers so far.

## To-do list

- [ ] This pull request is ready to merge. (of course I want to hear opinions first 😃 this is also my frist contribution so I'm not sure how the process is here)


fedeci (commented):
This is an interesting solution, but is too similar to the current `customValidator` validator. For full extensibility we have to inject validators and sanitizers directly in the chain.

ysfaran:
@fedeci could you please explain what you mean with "full extensibility"? What are the plans here?

I agree though that it's quite similar to `customValidator` and it probably would make sense to reuse some code here.

fedeci (commented):
Right now we do not inject the extended `ValidatorsImpl` in the middlewares like `check()`, `body()`... so they would not be available in the chain
```js
// e.g.
body('name').isFoo() // not available
```
Pluggability of validators is something that we need to think more accurately also considering typescript requirements and security (we can continue discuss about it in the linked issue).
For the moment I like the idea of allowing the user to set options in `.custom()` and `.customSanitizer()`, but I would prefer to keep the `meta` parameter as the second one and move `options` at the last position.

ysfaran:
I updated the PR and extended `CustomValidator` so it can work with custom options aswell.

> [...] so they would not be available in the chain

With the change I did they should be available actually. I created a fully working typescript example in this test project:  https://github.com/ysfaran/express-validator-extend-test

With `npm link` I was able to test everything locally, so let's have a look at the example.

In `extensions.ts` I created a custom validator `isToken(value, meta, options)` that optionally receives a `length` property as part of `options`. (it's worth noting that `options` could also be a simple `number` here). I am able to add this validator globally, because of `extend` which is part of this PR:

```ts
// extensions.ts
import { extend } from "express-validator";

type TokenValidatorOptions = {
  length?: number;
};

extend.validators({
  isToken(value: string, meta, { length = 42 }: TokenValidatorOptions = {}) {
    if (typeof value !== "string") {
      throw Error("value must be a string");
    }

    if (!value.startsWith("0x")) {
      throw Error("value  must start with '0x'");
    }

    if (value.length !== length) {
      throw Error(`value must be ${length} characters long`);
    }

    return true;
  },
});
```

`extension.ts` needs to be loaded in a top-level file so it's executed before a custom validator is actually used, which is why it's imported at the top of `index.ts`:

```ts
// index.ts
import "./extensions";

import express from "express";
import { body, validationResult } from "express-validator";

const app = express();
app.use(express.json());

const port = 8080;

app.post("/", body("token").isToken({ length: 3 }), (req, res) => {

  if (errors.isEmpty()) {
    res.send("body has indeed a valid token!");
  } else {
    res.status(400).send(errors.array());
  }
});

app.listen(port, () => {
  console.log(`server started at http://localhost:${port}`);
});
```

As you can see I use the newly added `isToken` validator here. The express server can be started with `npm start`.

I checked the functionality by executing following `curl` commands:

```bash
$ curl -H "Content-Type: application/json" -d "{\"token\":\"0x1\"}" http://localhost:8080
body has indeed a valid token!
```
```bash
$ curl -H "Content-Type: application/json" -d "{\"token\":\"0x\"}" http://localhost:8080
[{"value":"0x","msg":"value must be 3 characters long","param":"token","location":"body"}]
```
 So ` body("token").isToken({ length: 3 })` actually works here.
 
 > [...] also considering typescript requirements [...]
 
 To make this actually work in typescript you have to tell the compiler that `isToken` actually existis in the chain:
 
 ```ts
 // express-validator.d.ts
 import "express-validator";

declare module "express-validator" {
  interface ValidationChain {
    isToken: (options?: { length?: number }) => ValidationChain;
  }
}
 ```
 
 This approach is similiar to [expect.extend](https://jestjs.io/docs/expect#expectextendmatchers) from `jest`.
 
 Please let me know if I missed something here 😃

gustavohenke:
Worked on something myself that superseded this.
See #1209

PR #1094, opened 2021-09-28
docs: warn users about bad practice

### LINKED ISSUE #1087: Docs have an example that promotes bad practice
## Describe the bug

https://express-validator.github.io/docs/custom-validators-sanitizers.html

On this page there is an example where it is checked if an email is already in use by accessing a data repository from a validator. Practice like this is bad as you'll only get a validation error if your database connection is dead for example. Proper error handling in such cases could be problematic. Issue described here: 

https://stackoverflow.com/questions/69136203/service-data-layer-exception-handling-in-express-validator/69136384#69136384

Business logic related validation and error handling should happen in the service layer: 
https://stackoverflow.com/questions/16394296/in-which-layer-should-validation-be-performed

## Suggestion:

Add a note on the docs page of the implications of accessing the data layer from a validator.

hariprasadkc:
## Description
Fixes #1087 
 https://express-validator.github.io/docs/custom-validators-sanitizers.html
The above documentation promotes a couple of bad practices that might affect the quality of the documentation.

- Illustration of accessing data layer directly from validation layer
- Promoting the usage of unvalidated data

### DIFF
--- a/docs/feature-custom-validators-sanitizers.md
+++ b/docs/feature-custom-validators-sanitizers.md
@@ -62,6 +62,8 @@ app.post('/user', body('email').custom(isValidUser), (req, res) => {
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
+> **Note:** In the example above, validation might fail even due to issues with fetching User information. The implications of accessing the data layer during validation should be carefully considered.
+
 ### Example: checking if password confirmation matches password
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/feature-custom-validators-sanitizers.md
```diff
@@ -62,13 +62,16 @@ app.post('/user', body('email').custom(isValidUser), (req, res) => {
...
+> **Note:** In the example above, validation might fail even due to issues with fetching User information. The implications of accessing the data layer during validation should be carefully considered.
+
 ### Example: checking if password confirmation matches password
 
 ```js
 const { body } = require('express-validator');
 
 app.post(
   '/user',
+  body('password').isLength({ min: 5 }),
```
fedeci, line 74: This doesn't add that more layer of security, but we can't do anything for that right now.

fedeci (approved):
Thanks!

Issue #1085, opened 2021-09-02
How to use CheckSchema to validate isPostalCode


fngch:
Hi folks,
I am using checkSchema to validate request body, unsure how to define the  schema for those validators with locale, it is fine by using check(...).isPostalCode( the options automatedly popup here, but in the checkSchema,it  does not, and not accepted passing 'US', and did not find a way to reference to type PostalCodeLocale in the schema definition.

Any help would be appreciate.

Charles F

fedeci:
Hi there, have you tried passing the `options: 'YOUR_LOCALE'`?
I know it is really annoying but we are (slowly 😞) working to fix it, there is a PR for v7 if I am not wrong.
```js
checkSchema({
	foo: {
		isPostalCode: {
			options: 'US'
 		}
	}
})
```

fngch:
Tried it, works fine now, thank you very much!

PR #1095, opened 2021-10-02
schema usage according to documentation results in Error

fkarg:
Current schema validation usage according to [the feature example](https://express-validator.github.io/docs/schema-validation.html) of `isPostalCode` results in an error: 

```json
    "errors": {
        "message": "Invalid locale 'undefined'"
    }
```

According to #1085, a PR for v7 is on the way to improve this (?). Anyway, I'd appreciate it if you could update this in the documentation of v6 as well. Do tell me if I should PR to a different branch (`v6.13-features` maybe?) instead.


fedeci (approved):
It's correct, thanks!

PR #1097, opened 2021-10-05
refactor: remove the unnecessary code

gabrielsimongianotti:
## Description

remove line 11, because it is not necessary


fedeci:
As you can see from the tests it is necessary :)

gabrielsimongianotti:
thank you very much

PR #1098, opened 2021-10-05
refactor: remove import not used

gabrielsimongianotti:
## Description

remove unused and inexistent import on `src/index.ts`


gustavohenke (changes requested):
thanks, it might be unnecessary but I'd like to keep it anyhow as it helps determining where things are coming from.

Issue #1106, opened 2021-11-08
Bump validator to 13.7.0


tux-tn:
Hello guys,

Validator [13.7.0](https://github.com/validatorjs/validator.js/releases/tag/13.7.0) has been published last week and is fixing two ReDOS vulnerabilities in `rtrim` and `isMagnetURI` that may affect `express-validator`.

I can work on it if nobody is already preparing a PR

fedeci:
I was planning to do it two days ago, but I am super messed with Fig and school stuff and I have no time. Would love a PR!

tux-tn:
Sorry didn't get back to you! I will gladly take the issue if you didn't start working on it yet.

fedeci:
Thanks!

fedeci:
@tux-tn have you started implementing this one? I can do it today and tomorrow.

tux-tn:
Really sorry, i started working on it but didn't finish all the changes, work and personal stuff took all my time 😓  Feel free to take the PR. I don't want to delay this change.

fedeci:
👍

PR #1110, opened 2021-11-25
upgrade validator to 13.7.0

elkarimareda:
## Description

upgrade library validator to 13.7.0
problem vulnerability


fedeci:
I am aware of this vulnerability, we already have plans for a PR that would support all the new validators too (see #1106). Furthermore just updating the dep breaks the package as you can see from the tests.

PR #1112, opened 2021-11-30
fix some old node cannot chainning the message

EddyPermana22:
## Description

on some old node, the chainning message validator not working.
fix this by copy the object and assign message using object destruction

## To-do list


fedeci:
On which old node version in particular is this happening?

fedeci:
> On which old node version in particular is this happening?

Friendly ping @EddyPermana22.

PR #1113, opened 2021-12-01
ci: add old node versions to CI

fedeci:
## To-do list

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
       # a more recent Node.js version
       fail-fast: false
       matrix:
-        node: [12, 14, 16]
+        node: [8, 10, 12, 14, 16, 17]
     steps:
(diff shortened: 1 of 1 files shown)


fedeci:
Oh, the version of jest we are using is not supporting node 8, I will fix it.

gustavohenke (changes requested):
I don't think we need to support such old versions. They have reached EOL already.

~Also parts of the code may not work, e.g. https://github.com/express-validator/express-validator/releases/tag/v6.4.0~ this might be gibberish

fedeci:
We are still mentioning support for Node 8 both in the readme and in the package.json. I 100% agree we should remove it (maybe also Node 10), but we need to wait for a major to definitely cancel them.

gustavohenke:
Ouch this has totally been an oversight of mine. I don't think that it'd work on Node 8 anyway.

PR #1120, opened 2021-12-13
Add `validationResult()` for schema validation example

daenamkim:
## Description

Missing `validationResult()` usage could make someone hard to use express-validator with Schema Validation at the beginning so it would be better to add little more codes.

## To-do list

- [x] I have added tests for what I changed. => No need to test, it is just fixing document exmaple.


fedeci (approved):
Thanks!

PR #1121, opened 2021-12-13
feat(types): add `files` object

fncolon:
## Description

`files` commonly used for handling I/O files data in expressjs. `(req.files)`

Capture

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -11,7 +11,7 @@ export interface Request {
 
 export type Middleware = (req: Request, res: any, next: (err?: any) => void) => void;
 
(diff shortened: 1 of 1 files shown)


gustavohenke (commented):
I suppose it could be allowed for that specific function you tried using (`buildCheckFunction`), but `files` is not standardised, so I'm not too comfortable adding an integration with e.g. multer.

btw @fedeci, it's probably worth for us to think of a guide on how to validate files

fedeci:
@gustavohenke I 100% agree with you.

PR #1130, opened 2022-02-13
Feature/report failed conditions

### LINKED ISSUE #1124: Include which validator failed in errors.mapped()
## Is your feature request related to a problem? Please describe.
This is not a functionality issue - it's an ease of use enhancement for validation messages. Let me know if there is already a way to get the information I'm looking for.

Basically, I can't see what validator failed in the output from `errors.mapped()`. I can see that the validation failed, but all I get is "InvalidValue". I want to have expressive validation messages for every validator. I understand that express-validator does not want to be opinionated about what error messages are returned - that makes perfect sense. **My thought here is that if errors.mapped() returned the validator instance that failed, I'd be able to write a fairly simple generic middleware that would take that information and create a message in my message style.** But as far as I'm aware, there is no way to see *which* validation condition failed when looking at the output of errors.mapped().

I could add `withMessage` calls to my validators, but then I'm duplicating code in my validators which isn't great. For example, I'd need to write something like:

```js
body('balance').exists().withMessage('balance is required').isNumeric().withMessage('balance must be numeric').isLength({ min: 2, max: 10 }).withMessage('balance must be between 2 and 10 characters')
```

While that's doable for a single api parameter, writing and maintaining that for thousands of parameters across hundreds of api endpoints will almost certainly be a huge pain and end up out of sync (the validators will probably get updated and the messages will be forgotten and end up outdated).

## Describe the solution you'd like
_Note: code snippets and line numbers below were taken from my downloaded copy of the module from my node modules so they are the transpiled JS files_

Add a non-enumerable property called "condition" to the error output containing the `negated`, `validator`, and `options` properties from the contextItem that was being processed when the validation failed.

For example, for `StandardValidation` (src/context-items/standard-validation.js) line 14 where `context.addError` is called, we could add an additional parameter on `context.addError` to contain the validation condition that failed like so:

```js
if (this.negated ? result : !result) {
    context.addError(this.message, value, meta, this); // new 4th parameter `this`
}
```

Then, in `context.addError` (src/context.js) lines 64 and 72 where the errors are pushed to the `this._errors` array, we could add a nonenumerable property first to those array elements like so:

```js
    // Added new 4th parameter for contextItem
    addError(message, valueOrNestedErrors, meta, condition) {
        const msg = message || this.message || 'Invalid value';
        if (meta) {
            let errorToPush = {
                value: valueOrNestedErrors,
                msg: typeof msg === 'function' ? msg(valueOrNestedErrors, meta) : msg,
                param: meta.path,
                location: meta.location,
            };
            // Add nonenumerable property containing contextItem before pushing to array
            Object.defineProperty(errorToPush, 'condition', { value: condition });
            this._errors.push(errorToPush);
        }
        else {
            let errorToPush = {
                msg,
                param: '_error',
                nestedErrors: valueOrNestedErrors,
            };
            // Add nonenumerable property containing contextItem before pushing to array
            Object.defineProperty(errorToPush, 'condition', { value: condition });
            this._errors.push(errorToPush);
        }
    }
```

Of course, we'd need to get all of the other contextItem classes updated as well.

I've tested this modification and it results in `errors.mapped` returning the `condition` element in the result as a nonenumerable property. The advantage of the nonenumerable property is it will not be automatically serialized by express so this will not change the return values of existing APIs that use express-validator (except in the extremely specific and extremely unlikely case where someone else already created a parameter called `condition` as a patch on the error output from express-validator)

The benefit here is that then I can then wrap the validation process in a middleware in my APIs that looks at the error and constructs a message based on the `condition` which includes the validation function (and by extension, its name), any options passed to that validation function, and whether or not that validation function was negated. It's fairly trivial with that information to construct a good message automatically.

Also - one quick side note: I noticed that `exists()` is implemented as a CustomValidator rather than a StandardValidator - does anyone know why that is?
I'm also curious if someone who knows this codebase better can tell me what other files would need to be updated to consistently show this type of error information.

amagid:
## Description

Closes #1124 

Attaching failed validation condition on calls to `context.addError`. This PR adds a nonenumerable property `"condition"` to errors resulting from failed validation. This property is nonenumerable and therefore should not affect any existing express APIs that return output from express-validator. However, it can be used to determine _which_ validation condition failed by reading error output and checking the `condition` property directly. This change maintains express-validator's goal of being unopinionated with regard to error messages while also allowing users to automatically generate human-readable error messages if they want to.

For example, if an input fails an isLength validation, `contextError.condition.constructor.name === 'isLength'`, so I would be able to generate an error message referencing the specific condition that failed rather than a generic message.

_To preserve existing functionality with minimal impact, the `condition` property is always marked optional in the `addError` function signature._

## To-do list

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -1,4 +1,5 @@
 import { ReadonlyContext } from './context';
+import { ContextItem } from './context-items';
 
 export interface Request {
   [k: string]: any;
@@ -38,6 +39,7 @@ export type ValidationError =
       // These are optional so places don't need to define them, but can reference them
       location?: undefined;
       value?: undefined;
+      condition?: ContextItem;
     }
   | {
       location: Location;
@@ -46,6 +48,7 @@ export type ValidationError =
       msg: any;
       // This is optional so places don't need to define it, but can reference it
       nestedErrors?: unknown[];
+      condition?: ContextItem;
     };
 
 // Not using Symbol because of #813
--- a/src/context-items/custom-validation.spec.ts
+++ b/src/context-items/custom-validation.spec.ts
@@ -25,7 +25,7 @@ const createSyncTest = (options: { returnValue: any; addsError: boolean }) => as
   validator.mockReturnValue(options.returnValue);
   await validation.run(context, 'bar', meta);
   if (options.addsError) {
-    expect(context.addError).toHaveBeenCalledWith(validation.message, 'bar', meta);
+    expect(context.addError).toHaveBeenCalledWith(validation.message, 'bar', meta, validation);
   } else {
     expect(context.addError).not.toHaveBeenCalled();
   }
@@ -53,13 +53,7 @@ describe('when not negated', () => {
         throw new Error('boom');
       });
       await validation.run(context, 'bar', meta);
-      expect(context.addError).toHaveBeenCalledWith('nope', 'bar', meta);
-    });
-
-    it('adds error with validation message if validator returns a promise that rejects', async () => {
-      validator.mockRejectedValue('a bomb');
-      await validation.run(context, 'bar', meta);
-      expect(context.addError).toHaveBeenCalledWith('nope', 'bar', meta);
+      expect(context.addError).toHaveBeenCalledWith('nope', 'bar', meta, validation);
     });
   });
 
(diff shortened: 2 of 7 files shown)


gustavohenke:
Closing, since it's seen no activity in a while.

PR #1150, opened 2022-05-23
fix: correctly run `.matches` when passing regex object

### LINKED ISSUE #1127: .matches failing when passing regex instead of pattern
## Describe the bug

It appears that passing a regex causes the validation to fail sometimes

Here's the typings just to show that regexps are allowed:

image

## To Reproduce

**FAILS**

image

image

image

**FAILS**
image

**PASSES**
image

## Expected behavior

The validation should always pass or always fail

## Current behavior

The validation fails sometime (it shouldn't)

## Environment:

- Express-validator version: 6.13.0
- Express version: 4.17.1
- Node 14.18.1

tonysamperi:
Forcing pattern to be string, to cope with lib validator which doesn't reset the RegExp instance
Fixes #1127 

## Description

Fixes matches validation failing one other time, passing regex value with g flag.

I didn't add any tests, because it works exactly as before, but simply extracts the source when passing the regex.

### DIFF
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -337,7 +337,7 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
       validator.matches,
       ...(typeof pattern === 'string'
         ? [pattern, modifiers]
-        : [pattern.source, ...[new Set((modifiers || '') + pattern.flags)].join('')]),
+        : [pattern.source, [...new Set((modifiers || '') + pattern.flags)].join('')]),
     ]);
   }
 }
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -333,8 +333,7 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isWhitelisted, chars);
   }
   matches(pattern: RegExp | string, modifiers?: string) {
-    return this.addStandardValidation.apply([
-      this,
+    return this.addStandardValidation.apply(this, [
       validator.matches,
       ...(typeof pattern === 'string'
         ? [pattern, modifiers]
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -334,7 +334,7 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
   }
   matches(pattern: RegExp | string, modifiers?: string) {
     return this.addStandardValidation.apply([
-      null,
+      this,
       validator.matches,
       ...(typeof pattern === 'string'
         ? [pattern, modifiers]
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -333,10 +333,12 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isWhitelisted, chars);
   }
   matches(pattern: RegExp | string, modifiers?: string) {
-    return this.addStandardValidation(
+    return this.addStandardValidation.apply([
+      null,
       validator.matches,
-      typeof pattern === 'string' ? pattern : pattern.source,
-      modifiers,
-    );
+      ...(typeof pattern === 'string'
+        ? [pattern, modifiers]
+        : [pattern.source, ...[new Set((modifiers || '') + pattern.flags)].join('')]),
+    ]);
   }
 }
--- a/src/chain/validators-impl.ts
+++ b/src/chain/validators-impl.ts
@@ -333,6 +333,10 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isWhitelisted, chars);
   }
   matches(pattern: RegExp | string, modifiers?: string) {
-    return this.addStandardValidation(validator.matches, pattern, modifiers);
+    return this.addStandardValidation(
+      validator.matches,
+      typeof pattern === 'string' ? pattern : pattern.source,
+      modifiers,
+    );
   }
 }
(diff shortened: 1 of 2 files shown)


src/chain/validators-impl.ts
```diff
@@ -333,6 +333,10 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isWhitelisted, chars);
   }
   matches(pattern: RegExp | string, modifiers?: string) {
-    return this.addStandardValidation(validator.matches, pattern, modifiers);
+    return this.addStandardValidation(
+      validator.matches,
+      typeof pattern === 'string' ? pattern : pattern.source,
+      modifiers,
```
fedeci, line 339: Isn't this counter-intuitive?
I mean this means that a user specifying a regex object with flags need to also set the optional modifiers separately
e.g.
```js 
  .matches(/foo/g) // => becomes /foo/
  .matches(/foo/g, 'g') // => is correctly /foo/g
```

fedeci (commented):
Hey @tonysamperi thanks for the PR, I am unsure how this fixes the linked issue, validator.js correctly handles passed in regex objects ([ref](https://github.com/validatorjs/validator.js/blob/cfcf9113c69b97477c409909a8729fa8efa0d595/src/lib/matches.js#L1-L9))

src/chain/validators-impl.ts
tonysamperi: Hi, yes, I forgot about the flags...I'll fix it tomorrow...

tonysamperi:
> Hey @tonysamperi thanks for the PR, I am unsure how this fixes the linked issue, validator.js correctly handles passed in regex objects ([ref](https://github.com/validatorjs/validator.js/blob/cfcf9113c69b97477c409909a8729fa8efa0d595/src/lib/matches.js#L1-L9))

Hi Federico, it fixes the problem I discovered (yes, I'm KikoCosmetics) in #1127.
Basically when you pass a RegExp instance to express-validator, its instance gets preserved.
I noticed that in lib "validator", if pattern is string, a new RegExp gets instantiated, but if it's not (which is the case), it gets used as it is. Which means that after the first match, the lastIndex on the instance won't be 0, causing the second test to fail.
The idea of my fix was passing always a string, so that the test is done with a fresh RegExp instance every time.
This for sure is how express-validator should be intended to work. 
I don't know if the behaviour of lib "validator" should be considered faulty.

Let me know what you think...we can also chat in Italian if it helps to better understand each other :)

fedeci:
Don't worry I got that later, I totally missed that explanation in the issue! I've also already updated the validatorjs issue to ask them to start using `.match`. Thanks for the fix!

tonysamperi:
Hi @fedeci I just added a commit...
I had to do something to merge regex flags with additional modifiers...
That's the cleanest way I could think of...
How does it look?

image

gustavohenke (approved):
Looks pretty solid and flexible to me 🙂  happy if @fedeci is

src/chain/validators-impl.ts
```diff
@@ -333,6 +333,11 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
     return this.addStandardValidation(validator.isWhitelisted, chars);
   }
   matches(pattern: RegExp | string, modifiers?: string) {
-    return this.addStandardValidation(validator.matches, pattern, modifiers);
+    return this.addStandardValidation.apply(this, [
+      validator.matches,
+      ...(typeof pattern === 'string'
+        ? [pattern, modifiers]
+        : [pattern.source, ...[new Set((modifiers || '') + pattern.flags)].join('')]),
+    ]);
```
fedeci, line 341: Just a super nitpick, can we use a ternary expression or an if statement, instead of the apply? That's definitely correct but makes the code harder to read for new developers.

fedeci: And can you add a couple tests to check that the flags added after the regex are effectively passed to the `addStandatdValidation` function?

tonysamperi: No problem for the tests,
But removing the apply would mean duplicate entirely the call to addStandardValidation...
I'd say that new developers shouldn't be supposed to touch such a piece of code!
I was reluctant myself despite several years of experience! 😅
Also remember that if they fix validator library we can revert this change! 😉

fedeci: That's a nice opinion and I'm fine to keep it as is, I'll open a PR and push to get that merged in asap!

fedeci: > And can you add a couple tests to check that the flags added after the regex are effectively passed to the addStandatdValidation function?

An even better testing way would be to reproduce the same regex in the issue, calling the method twice and asserting that the validation is passed both times.

tonysamperi: Ok I'll add the tests tomorrow!
Thanks!!

fedeci: Great! Thank you!

tonysamperi: Good thing you asked for the tests...The spread was misplaced...

fedeci:
Has this been closed for any particular reason? I'm sorry for the long delay in re-reviewing this PR!

tonysamperi:
Hi I needed to delete the repo, I did some cleanup of my github account.
At this point maybe they'll fix on validator.js side and we won't need this!

tonysamperi:
I can create another if you guys want btw

fedeci:
I see it really though they will fix soon, they are likely slower then us because they tend to wait to merge all PRs at once. If you prefer I can create this PR from your commit again tomorrow (you will get the commit credits so it should be fine). Let me know!

tonysamperi:
Don't worry I'll do some magic and maybe it will let me reopen this (I don't know if I can trick github by using the same branch name 😂)

Bye!

tonysamperi:
Sadly, no, it won't let me reopen this PR. :(

PR #1156, opened 2022-06-09
fix: correctly run .matches when passing regex object

tonysamperi:
## Description

Forcing pattern to be string, to cope with lib validator which doesn't reset the RegExp instance
Fixes https://github.com/express-validator/express-validator/issues/1127
Workaround of a behaviour which should be fixed in validator.js

## To-do list


fedeci:
Merging and publishing this tomorrow morning! Sorry again for the delay!

PR #1180, opened 2022-10-09
checkSchema support for multiple custom validators/sanitisers

gustavohenke:
## Description

This PR addresses #552 by adding support for multiple custom validators/sanitisers to `checkSchema()`.

```ts
checkSchema({
	email: {
		isEmail: true,
		isAllowedDomain: {
			bail: true,
			custom: value => value.endsWith('@gmail.com')
		},
		removeEmailAttribute: {
			// converts e.g. john+doe@gmail.com to john@gmail.com
			customSanitizer: value => value.replace(/(\+.+)?@/, '@'),
		},
		isExistingEmail: {
			custom: value => Users.findByEmail(value).then(/* do something */),
		},
	},
});
```

These names aren't used for anything but to allow placing the validators/sanitisers in different places in the resulting chain.

This is an advantage over using an array with `custom`/ `customSanitizer`, which wouldn't be too flexible.
For example, the following would run the custom validators in sequence, which may not be desired:

```ts
checkSchema({
	email: {
		customSanitizer: { options: removeEmailAttribute },
		custom: [
			{ options: isAllowedDomain },
			{ options: isExistingEmail },
		]
	}
})
```

## To-do list

- [x] Add tests for new custom validator syntax
- [x] Add tests for new custom sanitiser syntax
- [x] Add tests for mixing custom validator/sanitiser syntax with standard syntax
- [x] Add TS tests with `@ts-expect-error` for mixing custom validator/sanitiser syntax with standard syntax
- [x] Update docs


fedeci:
Reviewing this later today!

docs/feature-schema-validation.md
```diff
@@ -12,7 +12,7 @@ Its syntax looks like this:
 ```js
 const { checkSchema, validationResult } = require('express-validator');
 app.put(
-  '/user/:id/password',
+  '/user/:id',
```
gustavohenke, line 15: Just so that it fits better with the validated schema below.

docs/feature-schema-validation.md
```diff
@@ -69,7 +64,7 @@ app.put(
       },
     },
     // Support if functionality in schemas
-    someField: {
+    age: {
```
gustavohenke, line 67: ditto

gustavohenke:
@fedeci do you have any thoughts?

src/middlewares/schema.ts
```diff
@@ -80,6 +100,24 @@ export type ParamSchema = InternalParamSchema & {
...
+  SanitizersSchema &
+  {
+    [K in T]?: K extends keyof BaseParamSchema
+      ? BaseParamSchema[K]
+      : K extends keyof ValidatorsSchema
+      ? ValidatorsSchema[K]
+      : K extends keyof SanitizersSchema
+      ? SanitizersSchema[K]
+      : CustomSanitizerSchemaOptions | CustomValidatorSchemaOptions;
+  };
```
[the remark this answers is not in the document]
gustavohenke: Thanks for the ideas!
Can't do much formatting because of prettier 😛 

Don't know if the performance wins will be obvious if I change things around. Maybe it's coincidentally at its optimal already 😄 

The other suggestions are probably not going to work.
`CustomSanitizerSchemaOptions | CustomValidatorSchemaOptions` represents the value really, not a dictionary of possible keys/values that `K` might be in, so I have to dismiss your suggestion :(

But thanks again!!!

PR #1182, opened 2022-10-15
fix(default): not working when chained with `.optional()`

### LINKED ISSUE #1057: `.default()` sanitizer not working when chained
## Describe the bug

I am trying to use the `default()` sanitizer to make my optional array field have a default value. However, when I chain the `default()` sanitizer with other checks, the `default()` sanitizer doesn't do anything. In order for it to work, I need to make another check which only has the `default()` sanitizer attached to it. Instead of `body("...").isArray().optional().default([])`, I have to split it up into `body("...").isArray().optional()` and `body("...").default([])`.

## To Reproduce

1. Create a new `npm` project with the following dependencies/dev dependencies:
```
@types/express@4.17.12
express-validator@6.12.0
express@4.17.1
typescript@4.3.4
```
2. Create a new typescript file with the following code:
```typescript
import express from "express";
import { body } from "express-validator";

const app = express();

app.use(express.json());

app.post(
  "/fails",
  body("children").isArray().optional().default([]),
  (req: express.Request) => {
    console.log(`/fails output: ${JSON.stringify(req.body, null, 4)}`);
  }
);

app.post(
  "/works",
  body("children").isArray().optional(),
  body("children").default([]),
  (req: express.Request) => {
    console.log(`/works output: ${JSON.stringify(req.body, null, 4)}`);
  }
);

app.listen(3000);
```
3. Run the server and send a `POST` request with an empty body to both `/fails` and `/works`.

## Expected behavior

When the server is run and an empty `POST` request is sent to both `/fails` and `/works`, the following should show up in the console:
```
/fails output: {
    "children": []
}
/works output: {
    "children": []
}
```

## Current behavior

When the server is run and an empty `POST` request is sent to both `/fails` and `/works`, the following currently happens:
```
/fails output: {}
/works output: {
    "children": []
}
```

## Express-validator version:

- Version:  6.12.0

lpizzinidev:
fixes #1057


gustavohenke (commented):
Since `default()` now has an effect beyond just being a sanitizer, I think it should probably be in `ContextHandler`, next to `optional()`:

https://github.com/express-validator/express-validator/blob/02b69d47659148f71640701bb9c863001239de00/src/chain/context-handler.ts#L45-L53

Then, when a context is having field instances selected, the default value is applied:

https://github.com/express-validator/express-validator/blob/02b69d47659148f71640701bb9c863001239de00/src/context.ts#L75-L77

WDYT, @lpizzinidev?

lpizzinidev:
@gustavohenke 👍 Ok, I will apply the changes.

gustavohenke (commented):
Hey @lpizzinidev, it just occurred to me that this is a breaking change.

Reason: 
`default`, as a sanitizer, can be placed at any point in the chain, and it only changes the result from that point onwards, which is pretty much the issue with https://github.com/express-validator/express-validator/issues/1057.

<details>
<summary>Example</summary>

Say that there's a requirement that a fruit value is provided, and it must be at least 5 chars long:

```js
// A sanitizer that changes a fruit to `null` if it's value is `apple`
const sanitizer = (fruit) => fruit === 'apple' ? null : fruit;

check('fruit').customSanitizer(sanitizer).default('banana').isLength({ min: 5 })
check('fruit').default('banana').customSanitizer(sanitizer).isLength({ min: 5 })
```

In the first chain above, if you don't pass a value to `fruit` or if you pass `apple`, the result will be the same - and the validation will pass.
In the second example, if you pass a value of `apple`, users will be puzzled as it's 5 chars in length, but it still fails -- because the default of `banana` is applied too early.

</details>

I still think that your change is a good one, but it will have to be part of v7 (see #990).
There might be more changes needed here though -- but rest assured I can implement them later 😄

lpizzinidev:
@gustavohenke 
No problem! Let me know if further changes are needed 👍

PR #1198, opened 2022-12-28
gh: upgrade checkout and setup-node actions

gustavohenke:
#1197 but applying that change to all jobs in the file 🤦


gustavohenke:
still need to await for https://github.com/coverallsapp/github-action/pull/139 to land in order for the deprecation warning to be gone though

PR #1199, opened 2022-12-30
Docusaurus v2

gustavohenke:
⚠️ This is a pretty large PR.

Replaces docusaurus v1 with v2. There a bunch of changes; the most significant ones is the design change, of course:

Screen Shot 2022-12-30 at 14 30 18

Furthermore, doc versions now need to contain _all_ its markdown files and sidebar files, instead of only those that changed.
This definitely added a bit of weight to the number of files changed in this PR.


gustavohenke:
Have published to https://express-validator.github.io ✅

PR #1204, opened 2023-01-17
Unknown field validation

gustavohenke:
## Description

This PR implements long awaited exact matching of requests, a.k.a unknown field validations.

It works by
1. grabbing all contexts and extracting the fields and locations that they validate; then
2. creating a recursive tree out of the known fields paths; for example, `foo.bar` becomes `{ foo: { bar: {} }`; then
3. recursively iterating over each field under each validated request location, seeing if they match a known field or not.

I made the conscious decision of _not_ including headers and cookies in the default validated locations.
The reason is a bit obvious: browsers _will_ add stuff you weren't expecting.

The API is very flexible too. All of the below are valid:

```js
// Only a single id param can exist
app.get('user/:id', checkExact(param('id').isInt()), getUserRoute);

// Only the email and password body properties can exist
app.post('login', checkExact([
	body('email').isEmail(),
	body('password').notEmpty()
], loginRoute);

// Same as above, but with a schema
app.post('login', checkExact(checkSchema({
	email: { isEmail: true },
	password: { notEmpty: true }
})), loginRoute);

// Same as above, but as separate middlewares
app.post('login', [
	body('email').isEmail(),
	body('password').notEmpty(),
	checkExact()
], loginRoute);
```

It's important to notice in the above that checkExact is always the last middleware, or one that wraps others.
That's because it will report fields from chains after it as unknown.

Note: **performance wise, this might be slow**, depending on what the request and the validation rules look like.

I'm not adding docs in this PR as I wish to revamp them before v7 is released.

## To-do list


src/field-selection.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+    });
+  });
+
+  // This one seems controversial.
+  // The nested property wouldn't pass validation - unless it's optional, in which case it's fair to
+  // argue that 'foo' should be selected?
+  it('does not select parent when only nested field is known and not an object', () => {
+    const req = { body: { foo: 'bla' } };
+    const instances = selectUnknownFields(req, ['foo.bar'], ['body']);
+    expect(instances).toHaveLength(0);
```
gustavohenke, line 272: Noteworthy. Let's see what happens after releasing this

PR #1212, opened 2023-02-15
feat: update to support validator 13.9.0

fedeci:
## To-do list

### DIFF
--- a/declarations/validator.d.ts
+++ b/declarations/validator.d.ts
@@ -98,7 +98,7 @@ declare module 'validator' {
   ): boolean;
   export function isLocale(str: string): boolean;
   export function isLowercase(str: string): boolean;
-  export function isLuhnValid(str: string): boolean;
+  export function isLuhnNumber(str: string): boolean;
   export function isMagnetURI(str: string): boolean;
   export function isMACAddress(
     str: string,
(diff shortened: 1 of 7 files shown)


src/chain/validators-impl.ts
```diff
@@ -86,8 +86,10 @@ export class ValidatorsImpl<Chain> implements Validators<Chain> {
   equals(comparison: string) {
     return this.addStandardValidation(validator.equals, comparison);
   }
-  isAfter(date?: string) {
-    return this.addStandardValidation(validator.isAfter, date);
+  isAfter(date?: string): Chain;
+  isAfter(options?: Options.IsAfterOptions): Chain;
+  isAfter(dateOrOptions?: string | Options.IsAfterOptions) {
+    return this.addStandardValidation(validator.isAfter, dateOrOptions);
```
gustavohenke, line 92: This (and `isISBN`) could/should be a single overload, with a parameter of union type instead.

From [TS handbook](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads):

> we can invoke it with strings or arrays. However, we can’t invoke it with a value that might be a string or an array, because TypeScript can only resolve a function call to a single overload:
> ...
> Always prefer parameters with union types instead of overloads when possible

I think that because of the above, usage might become picky for nothing:

```js
typeof date === 'string' ? chain.isAfter(date) : chain.isAfter(date)
```

declarations/validator.d.ts
```diff
@@ -6,6 +6,7 @@ declare module 'validator' {
   ): boolean;
   export function equals(str: string, comparison: string): boolean;
   export function isAfter(str: string, date?: string): boolean;
+  export function isAfter(str: string, options?: import('../src/options').IsAfterOptions): boolean; // overload
```
gustavohenke, line 9: Interesting. Whoever made this change to validator didn't think of porting it to `isBefore` 😛

gustavohenke (approved):
Conditional LGTM with the requested changes addressed

PR #1214, opened 2023-02-18
Request-level bail

### LINKED ISSUE #1100: Middleware fail fast
I managed to find a way to fail-fast the validation chain with bail, but what about failing fast the middleware list?

Say, I have 

```javascript
.post(
  '/somewhere',
  body('name').exists(),
  body('phoneNumber').exists(),
  handleStuff
)
```

If the name validation fails, I don't care to check phoneNumber anymore... Wasn't able to find a way to fail fast in this case. In the case above, if both validations fail, I will get errors for both of them.

gustavohenke:
## Description

Closes #1100

Implements an option to `.bail()` so that no further contexts will run if there are any errors.

Supports:
- [x] `oneOf`
- [x] `checkSchema`
- [x] `checkExact`

⚠️ This is a perf-loss for those using any of the above functions, as the chains now need to be awaited on, otherwise results will be unreliable.

Suppose that you have a request with 1k fields, one validator on each field, each taking 10ms to complete.
Also suppose that no fields present errors, so every validator runs.

On my machine, the time it takes for running all validation chains looks like this:
- **With request-level bail on every field:** 10s
- **With request-level bail on half of all fields:** 10s (no difference)
- **With request-level bail on a third of all fields:** ~7s (roughly 1/3 better than the above scenarios)
- **Without request-level bail:** ~100 ms.

## To-do list

### DIFF
--- a/src/chain/context-handler-impl.spec.ts
+++ b/src/chain/context-handler-impl.spec.ts
@@ -10,6 +10,7 @@ let contextHandler: ContextHandler<any>;
 beforeEach(() => {
   builder = new ContextBuilder();
   jest.spyOn(builder, 'setOptional');
+  jest.spyOn(builder, 'setRequestBail');
   jest.spyOn(builder, 'addItem');
 
   contextHandler = new ContextHandlerImpl(builder, {});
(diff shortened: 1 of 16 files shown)


src/middlewares/one-of.ts
```diff
@@ -62,12 +69,25 @@ export function oneOf(
   const middleware = async (req: InternalRequest, _res: any, next: (err?: any) => void) => {
     const surrogateContext = new ContextBuilder().addItem(dummyItem).build();
 
-    // Run each group of chains in parallel, and within each group, run each chain in parallel too.
```
gustavohenke, line 65: This is a perf loss, but I don't see any other way of implementing this.

src/chain/context-handler-impl.spec.ts
```diff
@@ -20,6 +21,17 @@ describe('#bail()', () => {
     contextHandler.bail();
     expect(builder.addItem).toHaveBeenCalledWith(new Bail());
   });
+
+  it('does not sets request bail if level is unset or set to chain', () => {
```
gustavohenke, line 25:
```suggestion
  it('does not set request bail if level is unset or set to chain', () => {
```

src/middlewares/one-of.ts
gustavohenke: Latest version still is a perf loss. I've added a summary of changes to the PR description.

PR #1215, opened 2023-02-21
Rename field using custom validator

anasshakil:
## Description
This allows to rename parameters using custom validator.

Example
```js
app.post(
    "/user",
    // Promise
    body("search").custom((value) => {
        return new Promise((resolve, reject) => {
            if (value.includes("@")) {
                return resolve("email");
            }
            resolve();
        });
    }),
    // Basic 
    check("search")
        .custom((value) => {
            if (value.includes("+")) {
                return "phone";
            }
            return false;
        })
        .withMessage("An error occurred"),
    (req, res) => {
        // Handle the request
    }
);
```
closes: #785 

## To-do list
- [ ] Wildcard support

### DIFF
--- a/src/context-items/custom-validation.ts
+++ b/src/context-items/custom-validation.ts
@@ -20,8 +20,8 @@ export class CustomValidation implements ContextItem {
         context.addError(this.message, value, meta);
       }
       // rename field if return type is string
-      if (typeof actualResult === "string") {
-          context.renameFieldInstance(actualResult, meta)
+      if (typeof actualResult === 'string') {
+        context.renameFieldInstance(actualResult, meta);
       }
     } catch (err) {
       if (this.negated) {
--- a/src/context-items/custom-validation.ts
+++ b/src/context-items/custom-validation.ts
@@ -19,6 +19,10 @@ export class CustomValidation implements ContextItem {
       if ((!isPromise && failed) || (isPromise && this.negated)) {
         context.addError(this.message, value, meta);
       }
+      // rename field if return type is string
+      if (typeof actualResult === "string") {
+          context.renameFieldInstance(actualResult, meta)
+      }
     } catch (err) {
       if (this.negated) {
         return;
--- a/src/context.ts
+++ b/src/context.ts
@@ -112,6 +112,13 @@ export class Context {
     if (!instance) {
       throw new Error('Attempt to rename field that did not pre-exist in context');
     }
+    const { originalPath } = instance;
+    if (this.fields.length !== 1) {
+      throw new Error('Attempt to rename multiple fields.');
+    }
+    if (/\.|\*/g.test(originalPath)) {
+      throw new Error('Attempt to rename a wildcard field');
+    }
     instance.path = newPath;
   }
 }
--- a/src/context.ts
+++ b/src/context.ts
@@ -106,6 +106,14 @@ export class Context {
       });
     }
   }
+  renameFieldInstance(newPath: string, meta: Meta) {
+    const { path, location } = meta;
+    const instance = this.dataMap.get(getDataMapKey(path, location));
+    if (!instance) {
+      throw new Error('Attempt to rename field that did not pre-exist in context');
+    }
+    instance.path = newPath;
+  }
 }
 
 export type ReadonlyContext = Pick<


src/context-items/custom-validation.ts
```diff
@@ -19,6 +19,10 @@ export class CustomValidation implements ContextItem {
       if ((!isPromise && failed) || (isPromise && this.negated)) {
         context.addError(this.message, value, meta);
       }
+      // rename field if return type is string
+      if (typeof actualResult === 'string') {
+        context.renameFieldInstance(actualResult, meta);
```
gustavohenke, line 24: May I ask why you thought of making use of the custom validators?

Changing the meaning of the return of a custom validator will be quite disruptive for everybody, and is certainly a breaking change.

Perhaps a new function in the chain would be a better idea?

src/context.ts
```diff
@@ -106,6 +106,21 @@ export class Context {
...
+    if (!instance) {
+      throw new Error('Attempt to rename field that did not pre-exist in context');
+    }
+    const { originalPath } = instance;
+    if (this.fields.length !== 1) {
+      throw new Error('Attempt to rename multiple fields.');
+    }
+    if (/\.|\*/g.test(originalPath)) {
+      throw new Error('Attempt to rename a wildcard field');
+    }
```
gustavohenke, line 121: Nice thinking here. I suppose that the callback should be able to handle the name to give to a field though -- as it has access to the full original path.

gustavohenke (changes requested):

src/context.ts
anasshakil: I had some success on renaming wildcard paths but need to do more work for wildcard support

src/chain/validators.ts
```diff
@@ -36,6 +36,14 @@ export interface Validators<Return> {
    */
   custom(validator: CustomValidator): Return;
 
+  /**
+   * Adds a field rename functionality to the validation chain.
+   *
+   * @param evaluator the custom evaluator based on the `CustomValidator`
+   * @returns the current validation chain
+   */
+  rename(evaluator: CustomValidator): Return;
```
gustavohenke, line 45: Nice, this is better!
A few comments here though:
1. This shouldn't stay in the `Validators` interface. `ContextHandler` sounds more like it.
2. Let's add a type specific for the rename callback, so that `CustomValidator` can evolve independently from the rename functionality.
3. WDYT of accepting a string value too, which works as a shortcut? For example, `rename('banana')` automatically renames that field to `banana`.

src/context-items/rename-field.ts
```diff
@@ -0,0 +1,28 @@
+import { CustomValidator, Meta } from '../base';
+import { Context } from '../context';
+import { ContextItem } from './context-item';
+
+export class RenameFieldContextItem implements ContextItem {
+  message: any;
```
gustavohenke, line 6: When you move `rename` out of `Validators`, I think that this won't be necessary anymore

src/context-items/rename-field.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  constructor(private readonly evaluator: CustomValidator, private readonly negated: boolean) {}
```
gustavohenke, line 8: Ditto for `negated`

gustavohenke (changes requested):
General question: should the old property be removed when renaming the item?

`renameFieldInstance` will work for setting the new property with the same value as the old field instance, but it won't touch the old one.

https://github.com/express-validator/express-validator/blob/eba0cfef613204f894c65daf0dd0db5c7eb32de1/src/chain/context-runner-impl.ts#L50-L54

anasshakil:
> General question: should the old property be removed when renaming the item?
> 
> `renameFieldInstance` will work for setting the new property with the same value as the old field instance, but it won't touch the old one.
> 
> https://github.com/express-validator/express-validator/blob/eba0cfef613204f894c65daf0dd0db5c7eb32de1/src/chain/context-runner-impl.ts#L50-L54

Thanks for pointing out. Renaming only `instance.path` was a bug prone strategy, since it breaks the `dataMap`. I have added `removeFieldInstance` to remove the current property and adds a new property with renamed parameter.

Issue #1218, opened 2023-03-25
'ko-kr' line missing in MobilePhoneLocale type


Yoowatney:
## Description
the "ko-KR" option is missing from MobilePhoneLocale type

## Solution
Add the following line : "| ko-KR".

PR #1219, opened 2023-03-25
add "ko-KR" in MobilePhoneLocale type

Yoowatney:
## Description

I have added the missing 'ko-KR' line. #1218 

## To-do list

+) If necessary, I can provide test code for adding 'ko-KR' line


src/options.ts
```diff
@@ -252,6 +252,7 @@ export type MobilePhoneLocale =
   | 'it-IT'
   | 'it-SM'
   | 'ja-JP'
+  | 'ko-KR'
```
gustavohenke, line 255: let's keep it ordered alphabetically please

gustavohenke (approved):
LGTM with the change requested below.

Please note that we're amidst the release of v7, so it might take a bit of time to release this change.

PR #1229, opened 2023-06-28
docs: prefer `const` over `let`

kerolloz:
## Description

A small change in the example to use `const` because the variable is not reassigned 

## To-do list

### DIFF
--- a/docs/guides/manually-running.md
+++ b/docs/guides/manually-running.md
@@ -30,7 +30,7 @@ const { validationResult, ValidationChain } = require('express-validator');
 // sequential processing, stops running validations chain if the previous one fails.
 const validate = validations => {
   return async (req, res, next) => {
-    for (let validation of validations) {
+    for (const validation of validations) {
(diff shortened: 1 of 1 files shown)


docs/guides/manually-running.md
```diff
@@ -30,7 +30,7 @@ const { validationResult, ValidationChain } = require('express-validator');
 // sequential processing, stops running validations chain if the previous one fails.
 const validate = validations => {
   return async (req, res, next) => {
-    for (let validation of validations) {
+    for (const validation of validations) {
```
gustavohenke, line 33: Can you do the same to the TypeScript counterpart of this code block?

gustavohenke (changes requested):

docs/guides/manually-running.md
kerolloz: Yup, sure.
Done ✅

gustavohenke (approved):

PR #1246, opened 2023-09-20
feat(matchedData): add generic types for return

otorrillas:
## Description

As brought up in https://github.com/express-validator/express-validator/issues/1041, it's currently _hard_ to type the result of `parsedData`. It comes with all fields set as `any` unless we do a [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) via the `as` keyword.
```ts
const { something } = parsedData(req);
// something has any type

const { something } = parsedData(req) as { something: string };
// something has string type
```

However, this syntax _overrides_ all type-checking. To mitigate that, this PR adds an optional [generic type](https://www.typescriptlang.org/docs/handbook/2/generics.html) to the `parsedData` function signature, so it becomes:

```ts
function matchedData<T = Record<string, any>>(
  req: Request,
  options: Partial<MatchedDataOptions> = {},
): T {
	// ...
}
```

Allowing users to keep using `matchedData` with the same signature as before:
```ts
const { something } = parsedData(req);
// something has any type
```

but to provide a type signature of the expected return (if it's been validated) if they'd like:
```ts
const { something } = parsedData<{ something: string }>(req);
// something has string type
```

There are **no breaking changes**, since `matchedData` still keeps `Record<string, any>` as the default return type.

> [!NOTE]
> While types might not be strictly accurate depending on how the generic type is created, I believe it's better than the current solution (`Record<string, any>`).
> As suggested in https://github.com/express-validator/express-validator/issues/1041#issuecomment-1150448639, though, I believe a more type could be provided from the library depending on the options provided (e.g. with `includeOptionals: true`) via [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types). I'm open to exploring that direction further if you (maintainers) believe it might be worth implementing 👍 

## To-do list

- [X] I have added tests for what I changed.

- [X] This pull request is ready to merge.

### DIFF
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -76,3 +76,35 @@ type Location = 'body' | 'cookies' | 'headers' | 'params' | 'query';
 
 Represents one of the request locations: `req.body`, `req.cookies`, `req.headers`, `req.params` and
 `req.query`.
+
+### `matchedData`
+
+The [`matchedData`](./matched-data.md) function signature accepts passing a [Generic Type](https://www.typescriptlang.org/docs/handbook/2/generics.html) as the return type.
+
+The default type is `Record<string, any>`.
+
+```ts
+import { matchedData } from 'express-validator';
+
+app.post(
+  '/contact-us',
+  [
+    body('email').isEmail(),
+    body('message').notEmpty(),
+    body('phone').optional().isMobilePhone()
+  ],
+  (req, res) => {
+    const result = validationResult(req);
+    if (!result.isEmpty()) {
+      // handle validation errors
+      return res.send('Please fix the request');
+    }
+
+    const result = matchedData<{
+      email: string,
+      message: string, 
+      phone?: string,
+    }>(req);
+  },
+);
+```
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -76,35 +76,3 @@ type Location = 'body' | 'cookies' | 'headers' | 'params' | 'query';
 
 Represents one of the request locations: `req.body`, `req.cookies`, `req.headers`, `req.params` and
 `req.query`.
-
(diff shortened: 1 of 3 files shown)


arafatkn (approved):
Looking good.

docs/api/misc.md
```diff
@@ -76,3 +76,35 @@ type Location = 'body' | 'cookies' | 'headers' | 'params' | 'query';
 
 Represents one of the request locations: `req.body`, `req.cookies`, `req.headers`, `req.params` and
 `req.query`.
+
+### `matchedData`
```
gustavohenke, line 80: Can you move these next to `matchedData`'s own documentation, in https://github.com/express-validator/express-validator/blob/master/docs/api/matched-data.md? Maybe under a "TypeScript usage" header.

Thanks!

gustavohenke (changes requested):

docs/api/misc.md
otorrillas: Sure thing! Thanks!

gustavohenke (approved):
thanks, this is an excellent enhancement 😃

src/matched-data.ts
```diff
@@ -49,8 +49,7 @@ export function matchedData(
     .filter(validityFilter)
     .map(field => field.instance)
     .filter(locationFilter)
-    .reduce((state, instance) => _.set(state, instance.path, instance.value), {})
-    .valueOf();
```
gustavohenke, line 53: Had to make a fix here; missed how [`_(...).reduce()` automatically unwraps the result value](https://lodash.com/docs/4.17.15#lodash), meaning that this `valueOf()` was incorrect here.

PR #1248, opened 2023-09-29
feat(ulid): ULID validation

arafatkn:
## Description

Added validation for ULID (Universally Unique Lexicographically Sortable Identifier).

Read more details on ULID in [github:ulid/spec](https://github.com/ulid/spec)
ULID Generator and validations on [https://ulid.page](https://ulid.page)

## To-do list
- [x] Updated readme

### DIFF
--- a/declarations/validator.d.ts
+++ b/declarations/validator.d.ts
@@ -151,6 +151,7 @@ declare module 'validator' {
   export function isTaxID(str: string, locale: import('../src/options').TaxIDLocale): boolean;
   export function isTime(str: string, options?: import('../src/options').IsTimeOptions): boolean;
   export function isURL(str: string, options?: import('../src/options').IsURLOptions): boolean;
+  export function isULID(str: string): boolean;
   export function isUUID(str: string, version?: import('../src/options').UUIDVersion): boolean;
   export function isUppercase(str: string): boolean;
   export function isVariableWidth(str: string): boolean;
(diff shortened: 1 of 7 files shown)


declarations/validator.d.ts
```diff
@@ -151,6 +151,7 @@ declare module 'validator' {
   export function isTaxID(str: string, locale: import('../src/options').TaxIDLocale): boolean;
   export function isTime(str: string, options?: import('../src/options').IsTimeOptions): boolean;
   export function isURL(str: string, options?: import('../src/options').IsURLOptions): boolean;
+  export function isULID(str: string): boolean;
```
gustavohenke, line 154: Hey, this should only be for validator.js types. It seems it doesn't have a `isULID` yet.

Did you mean to wait for https://github.com/validatorjs/validator.js/pull/2294 to be merged?

If you rather have it in express-validator ASAP, I think you can remove it from here and manually document it in https://github.com/express-validator/express-validator/blob/master/docs/api/validation-chain.md.

arafatkn: @gustavohenke Thanks for the review.
This PR has no dependency on https://github.com/validatorjs/validator.js/pull/2294 so i'm removing the lines according to your suggestion.

src/chain/validators.ts
```diff
@@ -89,6 +89,13 @@ export interface Validators<Return> {
    */
   isString(): Return;
 
+  /**
+   * Adds a validator to check if a value is a ULID.
+   *
+   * @returns the current validation chain
+   */
+  isULID(): Return;
```
fedeci, line 97:
```suggestion
  isULID(): Return; // TODO: use validatorjs implementation when PR will be merged
```

fedeci (changes requested):
Thanks for the PR @arafatkn! Can you please add a couple tests in the `validators-impl.spec.ts` file? If you need any help please ping me!

fedeci (approved):
Thanks! Sorry for late response

PR #1254, opened 2023-10-10
build: move build files to `lib/` folder

fedeci:


.gitignore
```diff
@@ -10,6 +10,7 @@ npm-debug.log*
 # Build artifacts
 /src/**/*.js
 /src/**/*.d.ts
+/lib/
```
gustavohenke, line 13: guessing the two above aren't needed anymore?

gustavohenke (approved):
Sure, but why? 😄

.gitignore
fedeci: Yeah, I am keeping them for "backwards problems". If we remove them now, existing contributors that have built the repo at least once may commit them by mistake. This isn't required but it could be handy to keep for a while😄

fedeci:
I don't know😄, yesterday I was doing some stuff on the repo in vscode and I was constantly opening the wrong file. I thought this could help to declutter the main folder. Wdyt?

gustavohenke:
Sounds good. Let's also add a TODO note to remove those in a couple of months then

PR #1261, opened 2023-11-09
docs(guide/manually-running) typescript example should use `ContextRunner`

dillonstreator:
## Description

The "manually running" guide's typescript example should use `ContextRunner` instead of `ValidationChain` as it is a more minimal interface required by the example helper.

## To-do list


gustavohenke (approved):
Hey, thanks for the suggestion. It's even documented this way 🤦

Issue #1273, opened 2024-01-26
How to keep keys for optional fields in matchedData?


stenneepro:
## Describe the bug

I want to keep keys for optional fields in request body in `matchedData` response.

## To Reproduce

Let's say we have user model.
```
interface User {
  firstName?: string;
  lastName?: string;
  email: string
}
```

In frontend, user removed first name. and other fields are not changed.
So frontend sends request with only changed fields.
```json
{"firstName": null}
```

In this case matchedData function returns empty JSON, but I would like to keep firstName key.
There is a parameter called includeOptional, but if we set that parameter to True, `matchedData` function returns JSON which contains both `firstName` and `lastName` as key.
```json
{
  "firstName": null,
  "lastName": null
}
```

Moreover, I think it's more accurate that if keys exist in both request and validation schema, it should exist in `matchedData` return value.

## Environment:

- Express-validator version: 7.0.1
- Express version: 4.18.2
- Node.js version: 20.11.0

Issue #1048, opened 2021-06-11
matchedData does not include nullable optionals


AndrewBarba:
## Describe the bug

There's currently no easy way to allow a property to be set to null, _and_ exclude any data that was not validated. A common pattern for us is to sanitize the request body post-validation using the `matchedData` function, for example:

```
// POST /user { username: "Andrew", age: null }

await validate(req, [
  body("username").isString(), 
  body("age").optional({ nullable: true }).isInt(),
  body("name").optional().isString()
])

req.body = matchedData(req, { locations: 'body' })

assert(req.body.age === null) // fails
```

I know `matchedData` provides the `includeOptionals` option but this option will take _all_ optionals and then set them to undefined, which is definitely not what we want.

As is - there is no way to get access to just the nullable properties which I consider a bug in matchedData as it did in fact match against a null proper and validated it according to `optional({ nullable: true })`

## Expected behavior

assert(req.body.age === null) // succeeds

## Current behavior

assert(req.body.age === null) // fails

## Express-validator version:

- Version: 6.x

herbertpimentel:
+1 I agree, a nullable value should be inclueded in the matchedData, by definition, the rule says it can be null, and if it is null. it matches the definitions

mrollinsiv:
+1

mrollinsiv:
Is there any update on this issue?

erikhh92:
Yup, waiting here too. The problem is with this kind of types:

`interface Body {
  age: number | null
}`

If i want to update a item that now has a value of age: 5 and make it null, i can't do this because the data come filtered without the null value, so the update item remains the same.

PR #1275, opened 2024-01-30
feat: ignoreUndefined in includeOptional

Vladyslav531:
## Description

Add `ignoreUndefined` to `includeOptionals` in `matchedData`

`ignoreUndefined` allows  user to pass optional values which aren't undefined like null and falsy values.
It keeps current behaviour and allow user to customize matchedData.

#1048 
#1175 
#1273 

## To-do list

- [ x] I have added tests for what I changed.

- [ x] This pull request is ready to merge.

### DIFF
--- a/src/context.spec.ts
+++ b/src/context.spec.ts
@@ -297,7 +297,7 @@ describe('#getData()', () => {
     context = new ContextBuilder().setOptional('undefined').build();
     context.addFieldInstances(data);
 
-    expect(context.getData({ requiredOnly: true })).toEqual([data[1]]);
+    expect(context.getData({ includeOptionals: false })).toEqual([data[1]]);
   });
 
   it('filters out undefineds and nulls when context optional = null', () => {
@@ -307,7 +307,7 @@ describe('#getData()', () => {
     context = new ContextBuilder().setOptional('null').build();
     context.addFieldInstances(data);
 
-    expect(context.getData({ requiredOnly: true })).toEqual([]);
+    expect(context.getData({ includeOptionals: false })).toEqual([]);
   });
 
   it('filters out falsies when context optional = falsy', () => {
@@ -318,7 +318,7 @@ describe('#getData()', () => {
     context = new ContextBuilder().setOptional('falsy').build();
     context.addFieldInstances(data);
 
-    expect(context.getData({ requiredOnly: true })).toEqual([]);
+    expect(context.getData({ includeOptionals: false })).toEqual([]);
   });
 
   describe('when same path occurs multiple times', () => {
--- a/src/context.ts
+++ b/src/context.ts
@@ -9,6 +9,7 @@ import {
   ValidationError,
 } from './base';
 import { ContextItem } from './context-items';
+import { IncludeOptionals } from './matched-data';
 
 function getDataMapKey(path: string, location: Location) {
   return `${location}:${path}`;
@@ -67,10 +68,10 @@ export class Context {
     readonly message?: any,
   ) {}
 
-  getData(options: { requiredOnly: boolean } = { requiredOnly: false }) {
+  getData(options: { includeOptionals: IncludeOptionals } = { includeOptionals: true }) {
     const { optional } = this;
     const checks =
-      options.requiredOnly && optional
+      !options.includeOptionals && optional
         ? [
             (value: any) => value !== undefined,
             (value: any) => (optional === 'null' ? value != null : true),
@@ -94,7 +95,13 @@ export class Context {
 
         return instances;
       })
-      .filter(instance => checks.every(check => check(instance.value)))
+      .filter(instance => {
+        if (options.includeOptionals === 'ignoreUndefined') {
+          return checks.every(check => check(instance.value)) && instance.value !== undefined
+        } else {
+          return checks.every(check => check(instance.value))
+        }
+      })
       .valueOf();
   }
 
--- a/src/matched-data.spec.ts
+++ b/src/matched-data.spec.ts
@@ -73,6 +73,25 @@ describe('when option includeOptionals is true', () => {
   });
 });
 
+describe('when option includeOptionals is ignoreUndefined ', () => {
+  it('returns object with optional data which is not undefined', done => {
+    const req = {
+      headers: { foo: '123', bar: null },
+    };
+
+    const middleware = check(['foo', 'bar', 'baz']).optional({ values: 'null' }).isInt();
+
+    middleware(req, {}, () => {
+      const data = matchedData(req, { includeOptionals: 'ignoreUndefined' });
+      expect(data).toHaveProperty('foo', '123');
+      expect(data).toHaveProperty('bar', null);
+      expect(data).not.toHaveProperty('baz');
+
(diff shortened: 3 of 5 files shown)


Vladyslav531:
@gustavohenke Could you check this PR please?

src/matched-data.ts
```diff
@@ -7,12 +7,21 @@ interface FieldInstanceBag {
   context: Context;
 }
 
+/**
+ * Defines if includes optional data or not.
+ *
+ * - false: not include optional data
+ * - true: include optional data
+ * - `ignoreUndefined`: include optional data which is not undefined like null, falsy values
+ */
```
fedeci, line 16:
```suggestion
 * - false: do not include optional data
 * - true: include optional data
 * - `ignoreUndefined`: include optional data except for undefined values
 */
```

src/context.ts
```diff
@@ -94,7 +95,13 @@ export class Context {
...
         return instances;
       })
-      .filter(instance => checks.every(check => check(instance.value)))
+      .filter(instance => {
+        if (options.includeOptionals === 'ignoreUndefined') {
+          return checks.every(check => check(instance.value)) && instance.value !== undefined;
+        } else {
+          return checks.every(check => check(instance.value));
+        }
+      })
```
fedeci, line 104: This piece of code should stay unedited and the update should be made upper in the various checks definition.

src/context.spec.ts
fedeci, line 1: Can you add some tests here to check if `includeOptionals: 'ignoreUndefined' works`

src/matched-data.ts
```diff
@@ -7,12 +7,21 @@ interface FieldInstanceBag {
...
+export type IncludeOptionals = boolean | 'ignoreUndefined';
```
fedeci, line 17: Can you move this definition to the `context.ts` file?

src/context.ts
```diff
@@ -67,10 +68,10 @@ export class Context {
     readonly message?: any,
   ) {}
 
-  getData(options: { requiredOnly: boolean } = { requiredOnly: false }) {
+  getData(options: { includeOptionals: IncludeOptionals } = { includeOptionals: true }) {
```
fedeci, line 71: This is a breaking change, the function signature should not be changed.

fedeci (changes requested):
Thanks! Looks like a good new feature, there are some changes to publish this in a minor version.

fedeci:
Also, docs should be updated too once we are good to go with the implementation and the new API.

src/context.ts
Vladyslav531: I am not sure why it is breaking change. Could you explain in detail?

Vladyslav531:
> Also, docs should be updated too once we are good to go with the implementation and the new API.

I update docs, you can check

src/context.ts
fedeci: Because the Context class is exported and exposed to the public, so changing the signature of the function (in this case the parameters) is a breaking change.

src/matched-data.ts
```diff
@@ -52,7 +55,7 @@ export function matchedData<T extends object = Record<string, any>>(
     .reduce((state, instance) => _.set(state, instance.path, instance.value), {} as T);
 }
 
-function createFieldExtractor(removeOptionals: boolean) {
+function createFieldExtractor(removeOptionals: IncludeOptionals) {
```
fedeci, line 58: This one is not a public function so you can change its signature.
I'll explain you why. Private functions are only used from the code inside the package, so we have full control over it. Instead public functions, classes, methods require extra caution when edited because we do not want to introduce a breaking change to someone using the library when we publish a minor version. When we publish a major version we can potentially do whatever we want with breaking changes.

src/context.spec.ts
```diff
@@ -310,6 +310,16 @@ describe('#getData()', () => {
     expect(context.getData({ requiredOnly: true })).toEqual([]);
   });
 
+  it('filters out undefineds only when includeOptionals is ignoreUndefined', () => {
```
fedeci, line 313:
```suggestion
  it('include null values when `ignoreUndefined` is used', () => {
```

src/context.ts
```diff
@@ -67,15 +76,17 @@ export class Context {
...
     const checks =
-      options.requiredOnly && optional
+      options.requiredOnly === true && optional
         ? [
             (value: any) => value !== undefined,
             (value: any) => (optional === 'null' ? value != null : true),
             (value: any) => (optional === 'falsy' ? value : true),
           ]
+        : options.requiredOnly === 'ignoreUndefined'
+        ? [(value: any) => value !== undefined]
```
fedeci, line 89: In this way we can manage to get a nice solution that also encourages developers to use right now the new api. In this way, internally we can still call getData() using includeOptionals (e.g. in context runner)
```suggestion

    // TODO(8.0.0): remove requiredOnly
    getData(options: {
        /**
        * It is ignored is `includeOptionals` is specified
        * @deprecated use `includeOptionals` instead
        */
        requiredOnly: boolean,
        includeOptionals: IncludeOptionals
    } = { requiredOnly: false }) {
        const { optional } = this;
        const includeOptionals: IncludeOptionals = includeOptionals !== undefined ? includeOptionals : !requiredOnly
        const checks = 
            !includeOptionals && optional
            ? [
                (value: any) => value !== undefined,
                (value: any) => (optional === 'null' ? value != null : true),
                (value: any) => (optional === 'falsy' ? value : true),
            ]
            : includeOptionals === 'ignoreUndefined' && optional
            ? [
                (value: any) => value !== undefined,
            ]
            : []
```

fedeci:
I also changed my mind on the naming `ignoreUndefined`. It may be confusing. What about `discardUndefined`?
/cc @gustavohenke

Vladyslav531:
> I also changed my mind on the naming of `ignoreUndefined`. It may be confusing. What about `discardUndefined`? /cc @gustavohenke

I think it is also good name.

src/matched-data.spec.ts
```diff
@@ -73,6 +73,25 @@ describe('when option includeOptionals is true', () => {
...
+
+    middleware(req, {}, () => {
+      const data = matchedData(req, { includeOptionals: 'discardUndefined' });
+      expect(data).toHaveProperty('foo', '123');
+      expect(data).toHaveProperty('bar', null);
+      expect(data).not.toHaveProperty('baz');
+
+      done();
+    });
+  });
```
fedeci, line 92:
```suggestion
  it('returns object with optional data which is not undefined', done => {
    const req = {
      headers: { foo: '123', bar: null, 'boo': false },
    };

    const middleware = check(['foo', 'bar', 'baz', 'boo']).optional({ values: 'null' }).isInt();

    middleware(req, {}, () => {
      const data = matchedData(req, { includeOptionals: 'discardUndefined' });
      expect(data).toHaveProperty('foo', '123');
      expect(data).toHaveProperty('bar', null);
      expect(data).toHaveProperty('boo', false);
      expect(data).not.toHaveProperty('baz');

      done();
    });
  });
```

fedeci (approved):

PR #1276, opened 2024-02-01
docs: typescript example should use isEmpty method (guide/manually-running)

DongSeonYoo:
## The example in the manual run guide in the official document does not work, 

The errors property of the Result<T> class in validation-result.ts is not accessible because it is a private property.
- I thought about opening the errors property as public, but I didn't think it was a good way to encapsulate it

## To-do list


gustavohenke (approved):
Fair call :)

PR #1279, opened 2024-02-16
fix: correctly select properties of primitives

### LINKED ISSUE #1245: Validation on nested fields not working.
## Describe the bug

For some inputs, validation on nested fields accessed by dot (.) is not returning error
This was working in v6 but changed in  v7

## To Reproduce

Please check https://runkit.com/tony-sunny/6503695da7894b0008762533

## Expected behavior

Expected validation chain to return error

## Current behavior

There are no errors in validaiton result

## Environment:

- Express-validator version:  7.0.1
- Express version: 4.18.2
- Node.js version: 18.16.1

fedeci:
## Description

Fixes #1245 
It is correct to assume that primitives will not have properties in a request.
Despite that express-validator should provide the correct result, so if a user wants to validate a non existent property we should treat that as undefined.
Eventually in the future we may also update `FieldInstance` to look something like this:

```typescript
export interface FieldInstance {
  path: string;
  originalPath: string;
  location: Location;
  value: any;
  exists: boolean // added property
}
```

## To-do list

### DIFF
--- a/src/express-validator.spec.ts
+++ b/src/express-validator.spec.ts
@@ -165,22 +165,3 @@ describe('ExpressValidator', () => {
     });
   });
 });
-
-describe('High level tests', () => {
-  const { body } = new ExpressValidator();
-  it('should error when .exists() is used on a nested field of a primitive', async () => {
-    const req = { body: { foo: 'hello' } };
-    const result = await body('foo.nop').exists().run(req);
-    expect(result.isEmpty()).toEqual(false);
-  });
-  it('should not error when .not().exists() is used on a nested field of a primitive', async () => {
-    const req = { body: { foo: 'hello' } };
-    const result = await body('foo.nop').not().exists().run(req);
-    expect(result.isEmpty()).toEqual(true);
-  });
-  it('should not error when .exists() is used on a nested field of a primitive using wildcard', async () => {
-    const req = { body: { foo: 'hello' } };
-    const result = await body('foo.*.nop').exists().run(req);
-    expect(result.isEmpty()).toEqual(true);
-  });
-});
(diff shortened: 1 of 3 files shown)


src/express-validator.spec.ts
fedeci, line 1: Do we have another place to store these tests?

src/field-selection.ts
```diff
@@ -52,13 +52,19 @@ function expandPath(object: any, path: string | string[], currPath: readonly str
...
+      return [];
     }
-
-    // there still are paths to traverse, but value is a primitive, stop
-    return [];
+    if (key === '*') {
+      // wildcard position does not exist
+      return [];
+    }
+    // value is a primitive, there are still still paths to traverse that will be likely undefined, we return the entire path
```
gustavohenke, line 66: Fun fact: you could theoretically go forever now (not for any good reason though), e.g. `foo.toFixed.name.toString.length.toFixed`... you just have to keep diving deeper into the prototype!

```suggestion
    // value is a primitive, paths being traversed from here might be in their prototype, return the entire path
```

src/field-selection.spec.ts
```diff
@@ -113,14 +115,19 @@ describe('selectFields()', () => {
     const req = {
       query: { foo: ['bar', 'baz'] },
     };
-    const instances = selectFields(req, ['foo[1]'], ['query']);
+    const instances = selectFields(req, ['foo[1]', 'foo[2]'], ['query']);
```
gustavohenke, line 118: OOC, does this cover anything non-obvious, or is just for thoroughness?

src/express-validator.spec.ts
gustavohenke: The tested aspects are the field selection + `exists` functionality.
I don't feel like they are needed though, everything is covered in some way already.

PR #1280, opened 2024-02-16
fix: validate entire array and not single elements

### LINKED ISSUE #1230: Empty array causing false positive
## Describe the bug

Hi!

When validations fields I noticed that array had a different behavior than other values.

If an empty array is given no validation is run and the field is considered valid even if we're not expecting an array value.
I think this line in [standard-validation.ts](https://github.com/express-validator/express-validator/blob/master/src/context-items/standard-validation.ts#L19) producing this behavior.

## To Reproduce

Here's a [RunKit](https://runkit.com/yatengaky/649ee98055b0b00008d2d17e) reproducing the error.

## Expected behavior

In the RunKit I'm expecting only `count1` and `count6` to be valid because the [doc](https://express-validator.github.io/docs/guides/validation-chain#standard-validatorssanitizers) states if value is an array that each element of an array is validated.

It seems a bit weird that all types fails except arrays which means most validators needs an extra `not().isArray()` to prevent unexpected values in following middlewares.

## Current behavior

In the RunKit , `count1`, `count5` and `count6` are valid.

## Environment:

- Express-validator version: 7.0.1
- Express version: 4.18.2
- Node.js version: 18.16.1

fedeci:
## Description

To be merged after #1279
Closes #1230
Closes https://github.com/express-validator/express-validator/issues/1243
Related https://github.com/express-validator/express-validator/issues/755#issuecomment-1933821161

Maybe it is not the best time for doing this since we just released v7, but I changed my mind about forcing the entire array to comply with the validator. 
We provide wildcards which are a great way to validate individual array items. 
This PR is born because there are a lot of cases where validating individual items if not requested by the user leads to unexpected results.
Otherwise we can: add a param to body(''), check('')... to explicitly enable/disable array, or keep the current approach.

@gustavohenke wdyt?

## To-do list

### DIFF
--- a/src/express-validator.spec.ts
+++ b/src/express-validator.spec.ts
@@ -183,4 +183,19 @@ describe('High level tests', () => {
     const result = await body('foo.*.nop').exists().run(req);
     expect(result.isEmpty()).toEqual(true);
   });
+  it('should error array if no wildcard is used', async () => {
+    const req = { body: { foo: ['a', 'b', 'd', 'c'] } };
+    const result = await body('foo').isIn(['a', 'd', 'b', 'c']).run(req);
+    expect(result.isEmpty()).toEqual(false);
+  });
+  it('should not error array if wildcard is used', async () => {
+    const req = { body: { foo: ['a', 'b', 'd', 'c'] } };
+    const result = await body('foo.*').isIn(['a', 'd', 'b', 'c']).run(req);
+    expect(result.isEmpty()).toEqual(true);
+  });
+  it('should error if only the first item matches', async () => {
+    const req = { body: { foo: ['a', 'b', 'd', 'c'] } };
+    const result = await body('foo').equals('a').run(req);
+    expect(result.array().length).toEqual(1);
+  });
 });
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "prepublishOnly": "tsc",
     "postpublish": "npm run docs:publish",
     "test": "jest",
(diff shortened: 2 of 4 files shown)


darkman82:
Hi,
currently, to avoid unexpected results with arrays passed when single value is expected, we are forced to:
- use isString() on each field for query string
- use isString() or not().isArray() for body according to expected field type

FYI: even the AI (ChatGPT) understood that a standard behavior should not include the use of that workaround on all fields,
as, for example, when expecting a single Boolean value in the body it suggests: v.body('active').isBoolean() while the current approach must be v.body('active').not().isArray().isBoolean.

darkman82:
Hi,
I think this is not a matter of training, even tough v6 and v7 share the same behavior regarding the by-pass problem,
it's the common sense that suggests that validators should expect single values unless specified otherwise. 

I discovered this problem by accident in 2019, after I had mistakenly duplicated the parameters in the query string. Then I realized that the issue was even more widespread, and obviously it crashed the entire logic.

If this was a wanted behavior, it is not obvious to me, nor to the AI.

src/express-validator.spec.ts
```diff
@@ -183,4 +183,19 @@ describe('High level tests', () => {
     const result = await body('foo.*.nop').exists().run(req);
     expect(result.isEmpty()).toEqual(true);
   });
+  it('should error array if no wildcard is used', async () => {
```
gustavohenke, line 186: Similar feeling to https://github.com/express-validator/express-validator/pull/1279#discussion_r1501438112.

gustavohenke (changes requested):

PR #1281, opened 2024-02-21
fix(checkExact): returns unknown nested fields under a wildcard

### LINKED ISSUE #1242: Is checkExact() with wildcards supported?
Hi all,

quick question: Is the checkExact() middleware supported for wildcards?

E.g. consider the following scenario:

I need to validate a body that looks something like this: 
```
const body = [
  {
    id: 1,
    qty: 100
  },
  {
    id: 2,
    qty: 100
  }
]
```
-> So essentially an array of objects with an id and qty property.

I can nicely check for validity with the following chain
```
  body()
    .isArray(),

  body("*")
    .isObject(),

  body("*.id")
    .isInt(),

  body("*.qty")
    .isInt(),
```

Adding `checkExact()` to the end of the chain above however does not seem to cause the following example to fail validation, despite it having one additional unwanted property (`wrong`) in the second object:

```
const body = [
  {
    id: 1,
    qty: 100
  },
  {
    id: 2,
    wrong: 1,
    qty: 100
  }
]
```

So I wonder: is this scenario even supported currently, or am I running into a bug (or maybe I am doing something wrong?)

thanks in advance

Kind regards,

Pano

fedeci:
## Description

Fixes #1242

## To-do list

### DIFF
--- a/src/field-selection.spec.ts
+++ b/src/field-selection.spec.ts
@@ -430,6 +430,49 @@ describe('selectUnknownFields()', () => {
     });
   });
 
+  it('selects unknown fields nested under a checked wildcard', () => {
+    const req = { body: { obj: { nestedobj1: { foo: true, bar: 123, boom: { foo: true } }, nestedobj2: { foo: true, baz: true } }, arr: [{ foo: 'a' }, { foo: 'a', bar: true }] } };
+    // arr and obj are checked
+    const instances = selectUnknownFields(req, ['obj', 'obj.*.foo', 'arr', 'arr.*.foo'], ['body']);
+    expect(instances).toHaveLength(4);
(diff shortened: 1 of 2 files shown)


gustavohenke (changes requested):
Smart way of fixing this, _but_ I'd like to discuss a gotcha with this.

A custom validator can validate any aspect of an object/array, which is an undocumented assumption that this PR breaks. It's arguable whether one could still want this behaviour when there's `isArray()`, `isObject()`, among others, nowadays.

Should we keep this behaviour?
If yes, we need to mark fields with any **user-specified custom validators** (which excludes e.g. `isArray()`/`isObject()`, implemented using `custom()`) as being considered known "as a whole", whereas others are "partially" known.

Either way, we need to document it.

Issue #1304, opened 2024-07-09
In custom() validator, please add option not to include its value in response


skysign:
## Is your feature request related to a problem? Please describe.
In ValidationChain, we might want to keep that certain query parameter is secured.
To secure query parameter, it should not be included HTTP response, even its custom validator is finalized as reject().

We can assume that the query parameter is API key.
If the API key is included in HTTP response, it can be written in a kind of log message.
It means that API key is exposed.

I mean API key query parameter should be handled like github action secret.

## Describe the solution you'd like
In ValidationChain, I hope to add new parameter in query() function.
To notice that the query parameter should be secured.

gustavohenke:
Hey, without looking at your code, can you take a look at `ValidationResult#formatWith` and let me know if that works for you?
Here's it in the docs: https://express-validator.github.io/docs/api/validation-result#formatwith

skysign:
Hi @gustavohenke 

Thanks for your comment.
I think that [formatwith](https://express-validator.github.io/docs/api/validation-result#formatwith) might be used for my problem.

However, I still think that this feature is good for express-validator.
To explain my issue, I've written example code below.

`
const router = express.Router()

async function myValidationResult(req: Request, res: Response, next: NextFunction) {
  const result = validationResult(req);
  if (!result.isEmpty()) {
    return res.status(400).json({ errors: result.array() })
  }

  return next();
}

function validationRules(): any[] {
  return [
    query('api_key', 'api_key is invalid.') <- Adding parameter here, to make api_key like github action secret
      .custom(async (api_key) => {
        const userldapDN = getLDAPDNfromAPIkey(api_key) <- This is our own logic for authentication
        const b = await IsValidAPIKey(api_key, userldapDN)

        return new Promise((resolve, reject) => {
          if (b) {
            resolve(b)
          }
          else {
            reject()
          }
        })
    })
  ]
}

router.get('/example-endpoint', validationRules(), myValidationResult, doExampleEndpoint)
`

Here is my example code to explain this issue.
we keep using myValidationResult() as there are a bit long validation rules.
Just we handle api_key something like github anction secret.

skysign:
It's very hard to write example code pretty.
Please see my gist below, to see my example code with pretty format.
https://gist.github.com/skysign/bf783d5df1b62260c5e0db0f3d119f6c

gustavohenke:
I see. It'd be a bit cumbersome to manually filter out using `formatWith`.

[...]

PR #1305, opened 2024-07-17
Add hide() method

skysign:
## Description

I'd like to contribute this commit.
This is an implementation for #1304.

## To-do list

- [X] This pull request is ready to merge.

### DIFF
--- a/docs/api/validation-chain.md
+++ b/docs/api/validation-chain.md
@@ -387,6 +387,74 @@ body('json_string').optional().isLength({ max: 100 }).isJSON().
 
 :::
 
+### `.hide()`
+
+```ts
+hide(fieldName: string, hiddenValue?: string): ValidationChain
+```
+
+Hide field's value in the returning errors of validationResult() method.
+The value might be confidential information (such as api key), so we replace its value.
+
+**Parameters:**
+
+| Name            | Description                                                                     |
+| --------------- | ------------------------------------------------------------------------------- |
+| `fieldName` | the field name which you want to hide, such as query parameter or cookie name and so on. |
+| `hiddenValue` | String to be replaced with field's value, if it's not given, by default, '********'. |
+
+:::info
+
+For example, we can implement custom api_key(in query parameter) validator, like below.
+
+```ts
+query('api_key', 'api_key is invalid.')
+  .custom(async (api_key) => {
+      const validity = await check_api_key(api_key);
+      return new Promise((resolve, reject) => { validity? resolve(validity): reject() }); })
+    })
+```
+
+Without, calling hide() method, if api_key is invalid,
+In the returning errors of validationResult() function,
+api_key query parameter is exposed.
+
+```ts
+{
+  "type": "field",
+  "value": "api key is exposed",
+  "msg": "api_key is invalid, please check that it is revoked.",
+  "path": "api_key",
+  "location": "query"
+}
+```
+
+However, api_key query parameter is hidden by calling hide() method, like below.
+
+```ts
+query('api_key', 'api_key is invalid.')
+  .custom(async (api_key) => {
+      const validity = await check_api_key(api_key);
+      return new Promise((resolve, reject) => { validity? resolve(validity): reject() }); })
+    })
+  .hide('api_key'),
+```
+
+The value of api_key query parameter is masked with '********'.
+As a result, it is not exposed.
+
+```ts
+{
+  "type": "field",
+  "value": "********",
+  "msg": "api_key is invalid, please check that it is revoked.",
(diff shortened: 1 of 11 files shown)


gustavohenke (commented):
General question about this PR before I make some nitpicks:
What do you think if the default behaviour was to simply omit the value in the errors, but if a hidden value is set, use it?

In other words, `query('api_key').hide()` would get reported as

```js
{ msg: 'Invalid value', path: 'api_key', location: 'query' }
```

whereas `query('api_key').hide('******')` would get reported as

```js
{ msg: 'Invalid value', path: 'api_key', location: 'query', value: '******' }
```

skysign:
> General question about this PR before I make some nitpicks: What do you think if the default behaviour was to simply omit the value in the errors, but if a hidden value is set, use it?
> 
> In other words, `query('api_key').hide()` would get reported as
> 
> ```js
> { msg: 'Invalid value', path: 'api_key', location: 'query' }
> ```
> 
> whereas `query('api_key').hide('******')` would get reported as
> 
> ```js
> { msg: 'Invalid value', path: 'api_key', location: 'query', value: '******' }
> ```

I think that it is a good idea, let me update my PR. :)

skysign:
As per your suggestion, the behavior of hide() method is updated.
Please review this PR. :)

skysign:
coverage is increased by adding new test case.
Please review this PR. :)

src/chain/context-handler.ts
```diff
@@ -107,4 +107,16 @@ export interface ContextHandler<Chain> {
         }
       | boolean,
   ): Chain;
+
+  /**
+   * Hide field's value in the returning errors of validationResult() method.
```
gustavohenke, line 112:
```suggestion
   * Hide the field's value in errors returned by `validationResult()`.
```

src/chain/context-handler.ts
```diff
@@ -107,4 +107,16 @@ export interface ContextHandler<Chain> {
...
+   *
+   * The value might be confidential information (such as api key), so we replace its value.
```
gustavohenke, line 114:
```suggestion
   * If the value is confidential information (such as api key),
   * you might want to call this method to prevent exposing it.
```

src/chain/context-handler.ts
```diff
@@ -107,4 +107,16 @@ export interface ContextHandler<Chain> {
...
+   *
+   * @param hiddenValue? String to be replaced with field's value
+   * - if it's not given, by default, '********'.
```
gustavohenke, line 117:
```suggestion
   * @param hiddenValue? String to replace the field's value with.
   *   If it's not set, the field value is removed from errors.
```

docs/api/validation-chain.md
```diff
@@ -387,6 +387,94 @@ body('json_string').optional().isLength({ max: 100 }).isJSON().
...
+Hide field's value in the returning errors of validationResult() method.
+If the value is confidential information (such as api key),
+please call this method to prevent exposing it.
+If hiddenValue is not given, value property itself is removed.
+
+| Name          | Description                               |
+| ------------- | ----------------------------------------- |
+| `hiddenValue` | String to be replaced with field's value. |
+
+:::info
```
gustavohenke, line 405: I'm not sure about the examples in this block.
Maybe we can get away with something like this?

```ts
// Omits the value in the errors
query('api_key').custom(isValidKey).hide(); 

// Replaces the value in the errors with '*****'
query('api_key').custom(isValidKey).hide('*****'); 
```

gustavohenke (changes requested):
Hey! I tried to make up for my delay in reviewing your PR, as you can see there are a few large changes 😅 

I think the docs need to have some of the suggestions applied, but use your judgement.

I also don't know yet what the behaviour should be for `oneOf()`, but it might be fine to ship this without support for `oneOf` for now.
WDYT?

skysign:
> Hey! I tried to make up for my delay in reviewing your PR, as you can see there are a few large changes 😅
> 
> I think the docs need to have some of the suggestions applied, but use your judgement.
> 
> I also don't know yet what the behaviour should be for `oneOf()`, but it might be fine to ship this without support for `oneOf` for now. WDYT?

Thanks for your suggestion regarding docs. Now, they are resolved.
I prefer that we may ship without support for `oneOf`.
How about to discuss in another PR regarding `oneOf`?

skysign (commented):
LGTM :)

gustavohenke (approved):

gustavohenke:
This PR has been released - check out https://github.com/express-validator/express-validator/releases/tag/v7.2.0 😃

PR #1344, opened 2025-11-10
Fix: isNumeric and other type validators incorrectly accepting arrays

### LINKED ISSUE #1323: `isNumeric` should reject for array
## Describe the bug

The `isNumeric` should throw the error message for the array but it does not

## To Reproduce

See this example code

```js
[
 checkSchema({
    id: {
      in: ['body'],
      exists: {
        errorMessage: 'ID is required',
      },
      isNumeric: {
        errorMessage: 'ID must be a number',
      },
    },
  }),
]
```

When I send a request with this body: 
```json
{
  "id": []
}
```

It sees the array as a number and then it passes the validation

## Expected behavior

IT should reject the request because the value is not a number

## Current behavior

IT should reject the request because the value is not a number

## Environment:

- Express-validator version: 7.2.0
- Express version: 4.21.2
- Node.js version: 20.15.1

djunehor:
## Description

This PR fixes #1323 where `isNumeric` and other type validators incorrectly accepted arrays (including empty arrays) instead of rejecting them. The issue occurred because empty arrays would skip validation entirely since the `forEach` loop had no elements to iterate over, causing the validation to incorrectly pass.

The fix adds array rejection logic to `StandardValidation` that checks if the validator is a type validator (like `isNumeric`, `isInt`, `isBoolean`, etc.) and rejects arrays before processing individual elements. This ensures that arrays are properly validated as invalid input for type validators since an array is not a number, string, or boolean.

Changes include:
- Added `shouldRejectArrays()` method to identify type validators that should reject arrays
- Modified `StandardValidation.run()` to reject arrays for type validators before element processing
- Updated existing tests to reflect the correct behavior (arrays should be rejected)
- Added comprehensive test coverage for array rejection scenarios

## To-do list


src/context-items/standard-validation.ts
```diff
@@ -24,4 +30,32 @@ export class StandardValidation implements ContextItem {
...
     });
   }
+
+  /**
+   * Determines if this validator should reject arrays entirely rather than validating individual elements.
+   * Type validators (numeric, string, etc.) should reject arrays since an array is not of those types.
+   */
+  private shouldRejectArrays(): boolean {
+    // Check if this is a type validator that should reject arrays
+    const typeValidators = [
```
gustavohenke, line 40: Doesn't this list pretty much every [standard validator](https://express-validator.github.io/docs/api/validation-chain#standard-validators)? 
WDYT of just making it a blanket case?

src/chain/validators-impl.spec.ts
```diff
@@ -157,16 +157,16 @@ describe('default #isBoolean()', () => {
     await isBoolean.run(context, 'false', meta);
     await isBoolean.run(context, 1, meta);
     await isBoolean.run(context, 0, meta);
-    // The two below are passing because we are using a StandardValidator which passes the value to `toString`
-    await isBoolean.run(context, [false], meta);
-    await isBoolean.run(context, ['true'], meta);
     expect(context.errors).toHaveLength(0);
 
+    // Arrays should now be rejected by type validators
```
gustavohenke, line 162: This comment would've made more sense as a note for reviewers - doesn't make sense when reading the code after merging

```suggestion
```

src/chain/validators-impl.spec.ts
```diff
@@ -228,9 +228,10 @@ describe('loose #isBoolean()', () => {
     await isBoolean.run(context, 1, meta);
     await isBoolean.run(context, 0, meta);
 
+    // Arrays should now be rejected by type validators
```
gustavohenke, line 231: ditto

```suggestion
```

src/context-items/standard-validation.ts
```diff
@@ -16,6 +16,12 @@ export class StandardValidation implements ContextItem {
   ) {}
 
   async run(context: Context, value: any, meta: Meta) {
+    // Check if this validator should reject arrays entirely
+    if (Array.isArray(value) && this.shouldRejectArrays()) {
```
gustavohenke, line 20: This is a breaking change, it'll certainly annoy people and we'll have to bring it back.

> The issue occurred because empty arrays would skip validation entirely since the forEach loop had no elements to iterate over, causing the validation to incorrectly pass.

This sounds like the core issue.
You can verify if the array is empty instead, then add the error.

gustavohenke (changes requested):

PR #1349, opened 2026-01-21
fix: skip optional validation for whitespace-only values

### LINKED ISSUE #1348: Optional fields with whitespace values still produce unclear validation errors
I noticed that there was a previously closed issue related to optional fields
and whitespace handling.

However, in current versions of express-validator, when optional() is used
together with validation chains (e.g. isInt()), fields that contain only
whitespace can still produce validation errors with messages that are not
clear for API developers.

This seems especially confusing when sanitizers like trim() are used before
or after optional().

Example:
body('age').optional().trim().isInt()

If the client sends:
age="   "

The resulting validation error does not clearly explain why the value was
considered invalid, even though the field is optional.

This is a common real-world API scenario. Improving error clarity or behavior
here would help developers avoid confusion.

KavyashreeSN:
Fixes #1348

This PR improves validation behavior for optional fields when values contain
only whitespace and are trimmed to an empty string.

When optional() is used, whitespace-only values are now skipped cleanly
instead of producing confusing validation errors. The change is minimal,
backward-compatible, and covered by targeted tests.

All existing tests pass and new tests cover the reported edge cases.


src/context.ts
```diff
@@ -90,10 +92,7 @@ export class Context {
       .flatMap((instances, group) => {
         const locations = _.uniqBy(instances, 'location');
 
-        // #331 - When multiple locations are involved, all of them must pass the validation.
-        // If none of the locations contain the field, we at least include one for error reporting.
-        // #458, #531 - Wildcards are an exception though: they may yield 0..* instances with different
-        // paths, so we may want to skip this filtering.
```
gustavohenke: This is important context to retain

PR #1353, opened 2026-03-13
fix: handle empty array validation in standard validators

### LINKED ISSUE #1243: Different behaviour v6/v7 using .notEmpty() to check empty array
## Describe the bug

We're using `.notEmpty()` to validate an incoming request body. After upgrading major version from 6 to 7, our unit tests started to fail. Turns out v7 `.notEmpty()` does not return a validation error for an empty array.

I do not see this in the breaking changes, unless it's part of some of the other points described there

## To Reproduce

Here you can see the different results:
1. Validation error for empty array on version 6.14.0: https://runkit.com/brianweet/express-validator-issue-template/6.14.0
2. No validation error on version 7.0.1: https://runkit.com/brianweet/express-validator-issue-template/7.0.1

## Expected behavior

v6 and v7 to have similar behaviour

## Current behavior

v7 does not return validation error for empty array

## Environment:

- Express-validator version: 7.0.1
- Express version: 
- Node.js version:

mahmoodhamdi:
## Problem

When using `.notEmpty()` on a field that contains an empty array `[]`, no validation error is produced. This is a regression from v6 behavior.

The root cause is in `StandardValidation.run()`:

```ts
const values = Array.isArray(value) ? value : [value];
values.forEach(value => { /* validate */ });
```

When `value = []`, the `forEach` iterates zero times, so validation is silently skipped. The empty array vacuously satisfies any validator, which is incorrect for validators like `notEmpty()` and `isEmpty()`.

## Solution

Changed the array check to also verify the array has items:

```ts
const values = Array.isArray(value) && value.length > 0 ? value : [value];
```

Empty arrays are now treated as a single value and go through the normal stringify path. `toString([])` returns `""`, which correctly evaluates as empty by `validator.isEmpty()`.

## Testing

Added a unit test in `standard-validation.spec.ts` that verifies empty arrays are validated as a single value instead of being skipped.

All existing 311 tests continue to pass, plus the 1 new test (312 total).

Fixes #1243

### DIFF
--- a/src/context-items/standard-validation.spec.ts
+++ b/src/context-items/standard-validation.spec.ts
@@ -55,6 +55,20 @@ it('calls the validator for each item in array field', async () => {
   expect(validator).toHaveBeenNthCalledWith(2, 'hey42');
 });
 
+it('validates empty array as a single value instead of skipping validation', async () => {
+  toString.mockImplementation(val => `hey${val}`);
+  validator.mockReturnValue(false);
+  await validation.run(context, [], meta);
+  expect(toString).toHaveBeenCalledWith([]);
+  expect(validator).toHaveBeenCalledWith('hey');
+  expect(context.addError).toHaveBeenCalledWith({
+    type: 'field',
+    message: validation.message,
+    value: [],
+    meta,
+  });
+});
+
 it('calls the validator with the value and options', async () => {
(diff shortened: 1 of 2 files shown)


gustavohenke (changes requested):
Thanks, but can you please update the docs around https://express-validator.github.io/docs/guides/validation-chain#standard-validatorssanitizers?

PR #1364, opened 2026-04-06
fix: handle read-only req properties when sanitizing (Express 5 compat)

### LINKED ISSUE #1325: Compatibility with Express 5: `req.query` is now read-only
Hi there! 👋 

Our project makes extensive use of express-validator for sanitizing query parameters and request bodies. We're currently migrating to **Express 5**, which as of version 5.1 has become the default for new projects.

According to the [Express 5 migration guide](https://expressjs.com/en/guide/migrating-5.html#req.query):

> The `req.query` property is no longer writable and is now a getter. 

This change prevents express-validator from mutating `req.query` directly, which breaks our existing sanitization setup.

This issue has been previously raised in [#1080](https://github.com/express-validator/express-validator/issues/1080) back in 2021, but it remains unresolved.

We’re looking for guidance on the recommended approach moving forward:
- Should we stop relying on express-validator for query sanitization?
- Is there a supported way to retrieve sanitized query values from express-validator?
- Are there any plans to adapt express-validator for Express 5 compatibility?

Thanks in advance for your help and for maintaining such a useful library!

Vansh1811:
## Description

Fixes #1325

Express 5 changed `req.query` to be a read-only property (a getter backed by a query parser, without a corresponding setter). When `express-validator` sanitizers run and write back the sanitized value via `_.set(req[location], path, newValue)`, a `TypeError: Cannot set property query of #<IncomingMessage>` which has only a getter is thrown.

This PR uses `Object.defineProperty()` to override the read-only property before writing back the sanitized value. This makes `express-validator` work gracefully with:

- **Express 4**: `req.query` is writable — sanitized values are written back as before
- **Express 5**: `req.query` is read-only — `Object.defineProperty()` makes it writable, then values are written back correctly

## Changes

- In `src/chain/context-runner-impl.ts`, used `Object.defineProperty()` to override read-only `req` properties (e.g. `req.query` in Express 5) before setting sanitized values, instead of silently ignoring `TypeError` exceptions

## To-do list

> Note: A proper test for this would require spinning up an Express 5 app in the test suite. If the maintainers prefer a specific test approach, I am happy to add it.

### DIFF
--- a/src/chain/context-runner-impl.ts
+++ b/src/chain/context-runner-impl.ts
@@ -28,6 +28,7 @@ export class ContextRunnerImpl implements ContextRunner {
     const bail = internalReq[contextsKey]?.some(
       context => context.bail && context.errors.length > 0,
     );
+
     if (bail) {
       return new ResultWithContextImpl(context);
     }
@@ -36,7 +37,6 @@ export class ContextRunnerImpl implements ContextRunner {
     context.addFieldInstances(instances);
 
     const haltedInstances = new Set<string>();
-
     for (const contextItem of context.stack) {
       const promises = context.getData({ requiredOnly: true }).map(async instance => {
         const { location, path } = instance;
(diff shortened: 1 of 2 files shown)


src/chain/context-runner-impl.ts
```diff
@@ -52,33 +52,38 @@ export class ContextRunnerImpl implements ContextRunner {
...
           // An instance is mutable, so if an item changed its value, there's no need to call getData again
           const newValue = instance.value;
-
           // Checks whether the value changed.
           // Avoids e.g. undefined values being set on the request if it didn't have the key initially.
           const reqValue = path !== '' ? _.get(req[location], path) : req[location];
           if (!options.dryRun && reqValue !== instance.value) {
-            path !== '' ? _.set(req[location], path, newValue) : _.set(req, location, newValue);
+            try {
+              // Express 5 makes req.query read-only (a getter without a setter).
```
gustavohenke: I'm not sure if we should try doing this, but you can override it if it's configurable with `Object.defineProperty()`.

WDYT?

gustavohenke (changes requested):
Please add some tests.

Vansh1811:
Added tests for the Express 5 read-only property handling in a new `describe('with read-only req properties (Express 5 compat)')` block in `context-runner-impl.spec.ts`:

1. **Does not throw when req location is read-only** — uses `Object.defineProperty` with a getter-only `req.query` to simulate Express 5, asserts the runner resolves without throwing.
2. **Still validates fields when req location is read-only** — confirms validation errors are still collected even when `_.set` silently fails.
3. **Rethrows non-TypeError exceptions** — asserts a `RangeError` thrown from a setter propagates up (only `TypeError` is silently ignored).

Also noted your inline suggestion about `Object.defineProperty()` — happy to explore that approach instead if you prefer it over the `try/catch`. Let me know!

Vansh1811:
Hi @gustavohenke, I've updated the PR with the Object.defineProperty() approach as you suggested in the review comments. Instead of wrapping the _.set() calls in a try/catch that silently swallows TypeError, the code now uses Object.defineProperty() to explicitly override the read-only property before setting it. This is more explicit and avoids silently ignoring errors.

The PR description and commit message have been updated to reflect this change. The existing tests still pass since they test the behavior, not the implementation details.

Please let me know if there are any other changes needed!

Vansh1811:
Hi @gustavohenke! Just checking in on this PR — I see there's 1 requested change. Could you point me to what needs to be updated? I'll also go ahead and rebase to bring the branch up to date. Happy to address any feedback to get this across the finish line. Thanks!

gustavohenke (changes requested):
Hey! Sorry for the delay. Incredibly busy year with almost no time left.

> Just checking in on this PR — I see there's 1 requested change. Could you point me to what needs to be updated?

You were probably looking at the previous review, which I think you addressed already.

Please take a look at the issues Copilot raised 🙇

PR #1369, opened 2026-05-05
fix: escape reconstructed field path segments

### LINKED ISSUE #1368: Wildcard field paths can overwrite values returned by matchedData()
Hi, I found a path reconstruction edge case where keys selected through a wildcard can be interpreted as lodash path syntax when `matchedData()` builds its return object.

### Reproduction

```js
const { body, matchedData, validationResult } = require("express-validator");

async function run(bodyObj) {
  const req = { body: bodyObj };

  await body("role").isIn(["user", "moderator"]).run(req);
  await body("*").isString().run(req);

  console.log(validationResult(req).array());
  console.log(matchedData(req));
}

run(JSON.parse("{\"role\":\"user\",\"[\\\"role\\\"]\":\"admin\"}"));
```

Current output on 7.3.2:

```js
[]
{ role: "admin" }
```

The `role` field itself passed the `isIn(["user", "moderator"])` check as `"user"`, but the wildcard-expanded key named `["role"]` is later passed to lodash `_.set()` as a path and overwrites the returned `role` value.

### Expected

`matchedData()` should preserve the literal wildcard-expanded key, for example:

```js
{ role: "user", "[\"role\"]": "admin" }
```

### Cause

`reconstructFieldPath()` currently only quotes segments containing a dot. Segments containing bracket syntax, quotes, or backslashes can still be reconstructed into strings that lodash treats as path syntax instead of literal field names.

This matters for apps that combine specific validators with wildcard sanitizers/validators and then trust `matchedData()` as the sanitized output.

I have a small PR ready that quotes those reconstructed segments so lodash keeps them as literal keys.

homanp:
Fixes #1368.

This updates `reconstructFieldPath()` so segments containing brackets, quotes, or backslashes are emitted as quoted bracket segments using `JSON.stringify()`. That keeps wildcard-expanded field names such as `["role"]` as literal keys when `matchedData()` later writes them with lodash `_.set()`.

The change preserves the existing behavior for normal object paths, numeric array indexes, and wildcard literals.

I also added regression coverage for:

- reconstructed paths containing brackets, quotes, and backslashes
- `matchedData()` keeping a wildcard-expanded `["role"]` key from overwriting a separately validated `role` field

Checks run:

- `npx jest src/field-selection.spec.ts src/matched-data.spec.ts --runInBand`
- `npm test -- --runInBand`
- `npm run build`
- `npm run lint`

### DIFF
--- a/src/field-selection.spec.ts
+++ b/src/field-selection.spec.ts
@@ -567,6 +567,9 @@ describe('reconstructFieldPath()', () => {
     ['numeric segment between text segments', ['foo', '0', 'bar'], 'foo[0].bar'],
     ['numeric segment followed by numeric segment', ['foo', '0', '0'], 'foo[0][0]'],
     ['text segment with a dot', ['foo', '.bar'], 'foo[".bar"]'],
+    ['text segment with brackets', ['foo', '["bar"]'], 'foo["[\\"bar\\"]"]'],
+    ['text segment with quotes', ['foo', 'bar"baz'], 'foo["bar\\"baz"]'],
+    ['text segment with backslashes', ['foo', 'bar\\baz'], 'foo["bar\\\\baz"]'],
   ])('%s', (_name, input, expected) => {
     expect(reconstructFieldPath(input)).toBe(expected);
   });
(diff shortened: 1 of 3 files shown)


gustavohenke (approved):
Thanks!

PR #1375, opened 2026-06-21
Use git+https repository metadata

Soengkit:
## Summary
- update the package repository metadata to use the public HTTPS GitHub URL
- keep the change limited to package metadata


gustavohenke (commented):
What does this do?

Soengkit:
Thanks for asking. This is a metadata-only change: it updates `package.json#repository.url` from the legacy `git://github.com/express-validator/express-validator.git` URL to `git+https://github.com/express-validator/express-validator.git`.

It does not change source code, build output, or runtime behavior. The intent is only to make the package metadata point at the public GitHub repository over HTTPS, so registry/tooling consumers do not rely on the unauthenticated `git://` protocol.

If you prefer plain `https://github.com/express-validator/express-validator.git` instead of the npm-style `git+https://` form, I can switch it.

gustavohenke:
> The intent is only to make the package metadata point at the public GitHub repository over HTTPS, so registry/tooling consumers do not rely on the unauthenticated `git://` protocol.

Ack - so this is the crux of your PR. It wasn't clear at all _why_ do this.

